PR #22, opened 2016-02-04
Update CoverageProviders and fix IdentifierResolutionMonitor

### LINKED ISSUE #21: Use Bibliographic Coverage Providers from core.
Change existing scripts and other code to use the Bibliographic Coverage Providers created in NYPL-Simplified/server_core#51 and NYPL-Simplified/circulation#59.

Currently there's the `OverdriveBibliographicMonitor` and the `ThreeMBibliographicMonitor` (at least) that have repeat functionality. This may require moving some things from circulation down to core, too.

courte:
This allows the IdentifierResolutionMonitor to interact with the new identifier-focused format of CoverageProviders to ensure coverage, updating the `ContentCafeCoverageProvider` and creating a coverage provider for the OA Content Server in the meantime.

Requires NYPL-Simplified/server_core#56. Fixes #21.

Next steps include:
- Confirming that the OCLC Coverage Providers function as expected
- Tossing all relevant Coverage Providers into the list
- Running UnresolvedIdentifiers through each provider individually and marking them presentation-ready as coverage is established.

### DIFF
--- a/content_cafe.py
+++ b/content_cafe.py
@@ -30,14 +30,14 @@ from core.util.summary import SummaryEvaluator
 class ContentCafeCoverageProvider(CoverageProvider):
     def __init__(self, _db):
         self._db = _db
+        self.input_identifier_types = [Identifier.ISBN]
+        self.output_source = DataSource.lookup(_db, DataSource.CONTENT_CAFE)
         self.mirror = ContentCafeCoverImageMirror(self._db)
         self.content_cafe = ContentCafeAPI(self._db, self.mirror)
-        input_identifier_types = [Identifier.ISBN]
-        output_source = DataSource.CONTENT_CAFE
 
         super(ContentCafeCoverageProvider, self).__init__(
             "Content Cafe Coverage Provider",
-            input_identifier_types, output_source,
+            self.input_identifier_types, self.output_source,
             workset_size=25)
 
     def process_item(self, identifier):
--- a/content_cafe.py
+++ b/content_cafe.py
@@ -27,18 +27,23 @@ from core.util.summary import SummaryEvaluator
 class ContentCafeCoverageProvider(CoverageProvider):
     def __init__(self, _db):
         self._db = _db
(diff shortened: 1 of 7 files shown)


content_cafe.py
```diff
@@ -27,18 +30,23 @@
 class ContentCafeCoverageProvider(CoverageProvider):
     def __init__(self, _db):
         self._db = _db
-        self.input_source = DataSource.lookup(_db, DataSource.CONTENT_CAFE)
-        self.output_source = self.input_source
+        self.input_identifier_types = [Identifier.ISBN]
```
leonardr: I think we don't actually want to get Content Cafe coverage for every single ISBN the metadata wrangler knows about. We only want to get coverage for ISBNs that a client specifically asks about. However, I think this code is fine because we're not going to actually run this like a normal coverage provider, we're going to use ensure_coverage on individual ISBNs.

content_server.py
```diff
@@ -0,0 +1,75 @@
...
+            Configuration.CONTENT_SERVER_INTEGRATION, required=True)
+        self.content_server = SimplifiedOPDSLookup(content_server_url)
+        self.importer = OPDSImporter(self._db, DataSource.OA_CONTENT_SERVER)
+        input_identifier_types = [Identifier.GUTENBERG_ID]
+        output_source = DataSource.OA_CONTENT_SERVER
+        super(ContentServerCoverageProvider, self).__init__(_db,
+                "OA Content Server Coverage Provider",
+                input_identifier_types, output_source, workset_size=10)
+
+    def process_item(self, identifier):
```
leonardr: Since this uses the lookup protocol you can do a batch of work at a time. Look at MetadataWranglerCoverageProvider in circulation/coverage.py. I bet you could even refactor a lot of code into a common subclass.

leonardr:
This looks good. My comments are more like things to think about for your next branches than reasons not to merge this branch.

courte:
Captured notes and next steps as issues: #23, #24, #25 & #26

Issue #23, opened 2016-02-05
Fix OCLC providers to work with new CoverageProvider code


courte:
They still depend on editions at the moment.

courte:
Fixed by #27.

PR #27, opened 2016-02-08
Fix oclc coverage providers

courte:
I selfishly made some white space changes & removed some unused libraries for this PR (sorry!) so it may be easier to walk through the commits themselves to see the changes. That said, this allows the existing OCLC Coverage Providers to work with identifiers and CoverageFailures, in accordance with the new CoverageProvider details in server_core.

I wanted to add some tests, but I don't see a clear way to do that without doing a good bit of refactoring at the moment, so I'll make an issue for the refactor itself.

### DIFF
--- a/gutenberg.py
+++ b/gutenberg.py
@@ -1,8 +1,14 @@
 import re
 from nose.tools import set_trace
 
-from core.coverage import CoverageProvider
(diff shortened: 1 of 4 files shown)


courte:
Fixes #23.

Issue #29, opened 2016-02-09
Make the OCLC API use the metadata layer


courte:
Currently, the OCLC API is passing around information using dictionaries. Force it to use the Metadata layer instead.

There's also a good chunk of code in the LinkedDataCoverageProvider that may be better suited to the API. :heavy_check_mark: it out.

PR #30, opened 2016-02-16
Add author to oclc data

courte:
I moved a bunch of methods previously in the coverage provider up into the API, and then I made sure that the creator viafs are being passed back to the Coverage Provider.

There's also a schweet script to pull the ThreeM books that are running about without authors and give them some contributors. Probably.

Paves the way for #29. Fixes NYPL-Simplified/server_core#17.

@leonardr: Sorry this PR is so big! I'll work on keeping them smaller & easier to review.

### DIFF
--- a/oclc.py
+++ b/oclc.py
@@ -98,6 +98,39 @@ class OCLCLinkedData(object):
     CAN_HANDLE = set([Identifier.OCLC_WORK, Identifier.OCLC_NUMBER,
                       Identifier.ISBN])
 
+    # We want to present metadata about a book independent of its
+    # format, and metadata from audio books usually contains
+    # information about the format.
+    UNUSED_TYPES = set([
+        'j.1:Audiobook',
+        'j.1:Compact_Cassette',
+        'j.1:Compact_Disc',
+        'j.2:Audiobook',
+        'j.2:Compact_Cassette',
+        'j.2:Compact_Disc',
+        'j.2:LP_record',
+        'schema:AudioObject',
+    ])
+
+    # Publishers who are known to publish related but irrelevant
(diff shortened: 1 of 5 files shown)


oclc.py
```diff
@@ -394,6 +431,199 @@ def extract_useful_data(cls, subgraph, book):
...
+            subgraph, publisher_uris)
+        publisher_names = [
+            i['schema:name'] for i in publishers
+            if 'schema:name' in i]
+        publisher_names = list(ldq.values(
+            ldq.restrict_to_language(publisher_names, 'en')))
+
+        for n in publisher_names:
+            if (n in self.PUBLISHER_BLACKLIST
+                or 'Audio' in n or 'Video' in n or 'n Tape' in n
```
leonardr: This is my old error, not yours, but I think 'n Tape' should just be 'Tape'.

oclc.py
```diff
@@ -1154,6 +1349,14 @@ def process_oclc_edition(self, original_identifier, edition):
             oclc_number.equivalent_to(
                 self.output_source, isbn_identifier, 1)
 
+        metadata = Metadata(self.output_source)
+        # Return contributor information.
+        VIAF_ID = re.compile("^http://viaf.org/viaf/([0-9]+)/$")
+        for viaf in edition['creator_viafs']:
+            viaf = VIAF_ID.search(viaf).groups()[0]
```
leonardr: There's a chance this regex won't match, because you put all viaf.org URLs into this list. With OCLC+LD this is the sort of problem that will happen one time in two million. Either use the same regex when putting URLs into this list down in the API, or skip a URL instead of processing it if search() returns None.

leonardr:
Looks good, just a couple minor things.

PR #31, opened 2016-02-19
Use metadata layer for oclcld

courte:
- Fixes the broken script from #30 (Thanks, @leonardr!)
- Refactors the OCLC-LD api to return a metadata object (Fixing #29). It's mimicking the Overdrive API pattern.
- Applies new metadata in the OCLC-LD coverage provider if an edition is being updated *or* new information is available from a new ISBN
- Merges new contributors to currently-authorless ThreeM books, thus fixing NYPL-Simplified/server_core#17.
- Has too many commits.
- Loves itself anyway.


oclc.py
```diff
@@ -575,11 +567,76 @@ def info_for_book_graph(self, subgraph, book):
...
+            if not isinstance(these_type_objs, list):
+                these_type_objs = [these_type_objs]
+            for this_type_obj in these_type_objs:
+                if isinstance(this_type_obj, dict):
+                    type_objs.append(this_type_obj)
+                elif isinstance(this_type_obj, basestring):
+                    type_objs.append({"@id": this_type_obj})
+        types = [i['@id'] for i in type_objs if
+                 i['@id'] not in self.UNUSED_TYPES]
+        return not not types
```
leonardr: ...This can't be right?

courte: >__> I'm forcing it to be a boolean. I guess I don't have to, though, since [] will be read as false in Python conditionals.

leonardr: How about `len(types) > 0`?

leonardr:
Overall this looks good! My only suggestion would be to make info_for_book_graph() the thing that makes a Metadata object. That way we skip the intermediate dictionary step. But you can do that after merging this branch, and maybe were planning on doing it anyway.

courte:
Ya, I went back and forth on that; there might actually be three methods that could be folded in. I just made an issue to do it. 

I also updated the bin/ script to be an executable and fixed that methods return value to be comprehensible, as suggested, in that last commit here. Merging momentarily.

PR #34, opened 2016-03-02
Make the identifier resolution process complete for ISBNs

leonardr:
This branch makes finalize() realize that you don't need to create a Work for an ISBN for the ISBN to be 'finalized' in a metadata wrangler sense.

I also change the identifiers_resolve script so that you can pass in some identifiers that need to be resolved (or made unresolved and then resolved). This makes problems easier to troubleshoot.

### DIFF
--- a/monitor.py
+++ b/monitor.py
@@ -40,6 +40,7 @@ from core.model import (
     DataSource,
     Edition,
     Equivalency,
+    Hyperlink,
     Identifier,
     LicensePool,
     Subject,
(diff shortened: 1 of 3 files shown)


monitor.py
```diff
@@ -160,6 +168,7 @@ def process_work(self, unresolved_identifier):
         license_pool = unresolved_identifier.identifier.licensed_through
         if license_pool:
             work, created = license_pool.calculate_work(even_if_no_author=True)
+        set_trace()
```
courte: You'll wanna remove this sucker.

leonardr: Thanks.

PR #36, opened 2016-03-03
Use metadata layer for oclcld

### LINKED ISSUE #32: Merge OCLC-LD api's formatting methods
OCLCLinkedData has two (or three) methods that turn raw data into some kind of holder object: tuple(`#extract_useful_data`), dict(`#info_for_book_graph`), and now Metadata(`#book_info_to_metadata`). Combine at least the last two. Determine whether the first should also be collapsed.

courte:
This combines the (now-dead) `#info_for_book_graph` method with the `#book_info_to_metadata` method. Super-cool.

It also fixes a couple parsing things to make sure we're pulling in all the subject information and VIAF ids we can handle.

My only concern is that the #extract_useful_data method doesn't appear to be pulling in publisher information or tags / genres anymore. I'll make an issue for it and follow-up.

Fixes #32.

### DIFF
--- a/oclc.py
+++ b/oclc.py
@@ -172,7 +172,6 @@ class OCLCLinkedData(object):
     ])
 
     FILTER_TAGS = POINTLESS_TAGS.union(TAGS_FOR_UNUSABLE_RECORDS)
-    UNUSABLE_RECORD = object()
 
     def __init__(self, _db):
         self._db = _db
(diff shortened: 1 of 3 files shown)


oclc.py
```diff
@@ -507,119 +558,34 @@ def info_for_book_graph(self, subgraph, book_info):
...
+        fixed_tags = [tag for tag in fixed if not None ]
+        subjects[Subject.TAG] = fixed_tags
+
+        for subject_type, subject_identifiers in subjects.items():
+            for subject_identifier in subject_identifiers:
+                metadata.subjects.append(SubjectData(
+                    type=subject_type, identifier=subject_identifier
+                ))
+
+        if (not metadata.links and not metadata.identifiers and
```
leonardr: I think finding some good contributor information (which you check for below) would be another reason to not return None.

PR #40, opened 2016-03-17
Organize bin

leonardr:
This branch organizes the `bin/` directory, creating a `repair/` subdirectory and moving into it a bunch of scripts intended for occasional ad hoc use rather than regular use driven by cron or Jenkins.

A lot of these scripts haven't been run in a while and no longer work, and I haven't tried to fix them beyond fixing the import code. My main goal here is to distinguish between the scripts that really need to be kept working and the ones where that's not as important.


bin/util/calculate_percentile
```diff
@@ -0,0 +1,33 @@
+# This is a utility script for calibrating a list of
```
courte: Does this need to be set as python?

PR #42, opened 2016-03-18
Clean up a bit

courte:
This is an old branch that started out as a desire to get rid of wierd-non-submodule-core and has become a combination of whitespace changes and deletions of unused libraries, etc.


leonardr:
I can't believe that weird-non-submodule-core is still in there.

Issue #41, opened 2016-03-17
Exception extracting subject name from OCLC-LD


leonardr:
While looking at the logs of the metadata wrangler's identifier resolution script I noticed an exception in OCLC's metadata extraction code:

```
2016-03-17 17:15:25,985:root:INFO:model.py:Cached http://www.worldcat.org/oclc/57657089.jsonld
2016-03-17 17:15:26,355:OCLC Linked Data Client:INFO:oclc.py:Processing edition None: None
2016-03-17 17:15:26,403:Identifier Resolution Manager:ERROR:monitor.py:FAILURE on Gutenberg ID/9536 ID=2421639 wr=267543 ("The Newcomes"): Traceback (most recent call last):
  File "/home/ec2-user/metadata/core/monitor.py", line 147, in resolve_and_handle_result
    self.resolve(unresolved_identifier)
  File "/home/ec2-user/metadata/core/monitor.py", line 183, in resolve
    self.finalize(unresolved_identifier)
  File "/home/ec2-user/metadata/monitor.py", line 153, in finalize
    self.resolve_equivalent_oclc_identifiers(identifier)
  File "/home/ec2-user/metadata/monitor.py", line 195, in resolve_equivalent_oclc_identifiers
    self.oclc_linked_data.ensure_coverage(oclc_id)
  File "/home/ec2-user/metadata/core/coverage.py", line 179, in ensure_coverage
    result = self.process_item(identifier)
  File "/home/ec2-user/metadata/oclc.py", line 1283, in process_item
    for metadata in self.api.info_for(identifier):
  File "/home/ec2-user/metadata/oclc.py", line 505, in info_for
    info = self.book_info_to_metadata(subgraph, book)
  File "/home/ec2-user/metadata/oclc.py", line 531, in book_info_to_metadata
    example_uris) = self.extract_useful_data(subgraph, book_info)
  File "/home/ec2-user/metadata/oclc.py", line 457, in extract_useful_data
    [subject_name] = name_value
ValueError: too many values to unpack
```
It doesn't happen every single time but it does happen pretty often. I don't see anything unusual in the OCLC-LD file that supposedly caused the problem, but I haven't looked into it too deeply.

courte:
Fun! I'll look into this.

courte:
There's also a recurring issues with `subject_data` not existing:

```
[...]

PR #43, opened 2016-03-21
Fix oclc extraction

courte:
This works to fix some of the recurring issues in #41 when an internal lookup isn't returned and when a subject has multiple potential names. I wasn't able to clearly replicate this issue, so I'm primarily going off the errors that are raised. I recreated those errors in the given test file (`tests/files/oclc/galapagos.jsonld`) by adding two subjects that don't have internal lookups: one that matches a known subject URI pattern and one that doesn't.

It also adds some logging to make the relevant cases easier to find in the future.

### DIFF
--- a/tests/test_oclc.py
+++ b/tests/test_oclc.py
@@ -386,11 +386,19 @@ class TestOCLCLinkedData(TestParser):
         eq_(u"11866009", oclc_id)
         eq_([u"Galápagos : a novel"], titles)
         eq_(1, len(descriptions))
+
+        # Even though there are 11 links in the books "about" list,
+        # "http://subject.example.wo/internal_lookup" does not get included as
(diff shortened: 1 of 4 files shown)


tests/test_oclc.py
```diff
@@ -392,7 +392,7 @@ def test_extract_useful_data(self):
         eq_(1, len(subjects[Subject.FAST]))
         eq_(4, len(subjects[Subject.TAG]))
         eq_(1, len(subjects[Subject.PLACE]))
-        eq_(2, len(subjects[Subject.LCSH]))
+        eq_(3, len(subjects[Subject.LCSH]))
```
leonardr: Stick a comment here saying that http://subject.example.wo/internal_lookup didn't get included because we couldn't find any information about it.

PR #44, opened 2016-03-25
Presentation policy

leonardr:
This branch brings the metadata wrangler up-to-date with the changes in https://github.com/NYPL-Simplified/server_core/pull/132.

* The code that derives a measurement of popularity from the number of editions of a book OCLC knows about has been moved from core into LinkedDataCoverageProvider.process_item. I'm not 100% sure this is the right place for this but I think so.

* The WorkPresentationCalculationScript has been removed in favor of the PresentationCalculationScript in core.


courte:
:+1:

PR #49, opened 2016-04-15
Enable authenticated collections

courte:
This works with NYPL-Simplified/server_core#183 to authenticate collections and track their lookups.


app.py
```diff
@@ -39,13 +40,27 @@ def initialize(cls, _db):
...
     Conf.initialize(_db)
 
+def accepts_collection(f):
+    @wraps(f)
+    def decorated(*args, **kwargs):
+        collection = CollectionController(Conf.db).authenticated_collection_from_request()
+        if isinstance(collection, Response):
+            return collection
+        elif collection:
+            return f(collection=collection, *args, **kwargs)
```
courte: This isn't quite how `circulation` manages authentication -- instead of passing the patron into certain methods, it requests authentication details again inside the method itself. But because the code is arranged slightly differently and many of the controllers here are shared, this feels "good enough for now" without rearranging a bunch of the Metadata Wrangler's controllers.

leonardr:
This looks good given that you make the appropriate changes in core.

bin/util/generate_collection
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env python
+"""Creates a new Collection with a provided name"""
+import os
+import sys
+bin_dir = os.path.split(__file__)[0]
+package_dir = os.path.join(bin_dir, "..", "..")
+sys.path.append(os.path.abspath(package_dir))
+from scripts import CollectionGeneratorScript
+
+CollectionGeneratorScript().run(sys.argv[1:])
```
daryanypl: Could you add a "correct way to call this is:" message?

aslagle:
:+1:

PR #52, opened 2016-04-25
Show collection updates

### LINKED ISSUE #51: Create a route for recent updates to metadata for a particular collection

courte:
This branch adds some tests for collection authentication -- both when it's optional and when it's required. It also gives the wrangler a new route that collections can hit, which will return only those identifiers updated since their last visit.

Fixes #51.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -40,7 +40,6 @@ class CollectionController(object):
         if isinstance(collection, ProblemDetail):
             return collection
 
-        # Record time of update check before initiating database query.
         last_update_time = flask.request.args.get('last_update_time', None)
         if last_update_time:
             last_update_time = datetime.strptime(last_update_time, "%Y-%m-%dT%H:%M:%SZ")
@@ -49,8 +48,10 @@ class CollectionController(object):
         pagination = load_pagination_from_request()
         works = pagination.apply(updated_works).all()
         title = "%s Updates" % collection.name
-        def update_url(page=None):
+        def update_url(time=last_update_time, page=None):
             kw = dict(_external=True)
+            if time:
+                kw.update({'last_update_time' : last_update_time})
             if page:
                 kw.update(page.items())
             return cdn_url_for("updates", **kw)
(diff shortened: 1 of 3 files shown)


leonardr:
:+1: pending changes to the core branch

leonardr:
:+1:

aslagle:
Isn't it a problem that last_checked is changed when you get one page? How do you get the next page?

courte:
@aslagle u rite, u rite. I'll reconsider.

leonardr:
Amy's absolutely right. Some part of me was thinking that the _client_ would be in charge of managing last_checked, so it didn't register that the server was also tracking it.

I do think the client needs to provide cutoff timestamps, because there's no guarantee that the client won't crash in the middle of its work. This is what Overdrive does for their analogous service.

controller.py
```diff
@@ -0,0 +1,75 @@
...
+            # (i.e. URN lookup) return None instead of an error.
+            return None
+        return INVALID_CREDENTIALS
+
+    def updates_feed(self):
+        collection = self.authenticated_collection_from_request()
+        if isinstance(collection, ProblemDetail):
+            return collection
+
+        # Record time of update check before initiating database query.
```
aslagle: this comment is wrong now

controller.py
```diff
@@ -0,0 +1,75 @@
...
+            last_update_time = datetime.strptime(last_update_time, "%Y-%m-%dT%H:%M:%SZ")
+        updated_works = collection.works_updated_since(self._db, last_update_time)
+
+        pagination = load_pagination_from_request()
+        works = pagination.apply(updated_works).all()
+        title = "%s Updates" % collection.name
+        def update_url(page=None):
+            kw = dict(_external=True)
+            if page:
+                kw.update(page.items())
```
aslagle: should the update url have the time in it, if there was one in the request?

courte:
Thanks, @aslagle! Fixed the things you commented on.

aslagle:
:+1:

PR #55, opened 2016-04-27
Remove ids from collections

### LINKED ISSUE #54: Add route for removing identifiers from collection

courte:
This branch allows a collection client to remove identifiers from its catalog as they are no longer included. It also moves the Wrangler slightly closer to the Circulation Manager's way of doing things, including using some of its bits, newly transferred to core: NYPL-Simplified/server_core#200.

Fixes #54.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -1,6 +1,6 @@
-import flask
 from nose.tools import set_trace
 from datetime import datetime
+from flask import request, make_response
 
 from core.app_server import (
     cdn_url_for,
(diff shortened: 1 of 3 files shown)


controller.py
```diff
@@ -74,3 +80,24 @@ def update_url(time=last_update_time, page=None):
             )
 
         return feed_response(update_feed)
+
+    def remove_items(self):
```
leonardr: I don't know how big a deal this is, but it seems weird that you might successfully remove some items from your collection and then get an error message because of invalid URNs. 

In the `lookup` handler I resolve this dilemma (you gave a bunch of URNs, some of them worked, some didn't) by serving an OPDS feed where every entry has a `message`. The message might be 'success!' or it might be 'invalid URN'. How much work would it be to do things that way here?

courte: Not much! I'll work on it.

leonardr:
Looks awesome. :+1:

PR #56, opened 2016-05-04
Remove unused test files

courte:
I'm removing a bunch of duplicated tests files, likely left over from when wrangler & core were but one. Many of these files have already been moved to circulation.

I left the content_cafe test files, even though they're not currently being used, because the Content Cafe integration is still being handled on metadata.


leonardr:
:x: :+1:

PR #57, opened 2016-05-05
Bring metadata up to date with latest core

leonardr:
This changes the metadata test runner to use the new test setup system, and improves a test that will otherwise fail once the licensepool-superceded core branch lands.


courte:
:+1:

PR #59, opened 2016-05-05
Use NoveList API

### LINKED ISSUE #53: Import equivalent identifiers from NoveList

courte:
This branch creates a service object for the NoveList Select API, and pulls in basic metadata. It also removes that annoying relative import thing! 🙌 

A couple lingering thoughts:
- I factored out the sample_data method in a really hack-y way. I want to move it into DatabaseTest, since it's used in all of our test directories, but I didn't want to spend too much time right now trying to figure out the path-related work that would need to happen.
- Their contributor data doesn't look great, and we could probably ignore it. (They return a single author's sort name. Even if there are multiple authors.)
- ~~I'm ignoring the NoveList unique identifiers, since ISBN's are more universal. Right now, this may mean duplicated work if multiple equivalent ISBNs seeks coverage.~~ See comment below.

Fixes #53. Depends on NYPL-Simplified/server_core#220.


novelist.py
```diff
@@ -0,0 +1,187 @@
...
+        self._db = _db
+        self.profile = profile
+        self.password = password
+        self.source = DataSource.lookup(self._db, DataSource.NOVELIST)
+
+    def lookup(self, identifier):
+        """Returns NoveList metadata for a particular identifier"""
+
+        client_identifier = identifier.urn
+        if identifier.type != Identifier.ISBN:
```
leonardr: I'd prefer to refuse to look up a non-ISBN identifier rather than try to find an equivalent. I think I know why you're doing this though. Let's talk it over.

novelist.py
```diff
@@ -0,0 +1,187 @@
...
+            appeals_info = feature_content.get('Appeals')
+
+        if series_info:
+            metadata.series = series_info['full_title']
+
+        if appeals_info:
+            extracted_genres = False
+            for appeal in appeals_info:
+                genres = appeal.get('genres')
+                if genres:
```
leonardr: What are the genres? Is there a small enough number of genres that it's worth creating a new Subject type with its own Classifier, rather than using `tag`?

novelist.py
```diff
@@ -0,0 +1,187 @@
...
+
+        if goodreads_info:
+            metadata.measurements.append(MeasurementData(
+                Measurement.RATING, goodreads_info['average_rating']
+            ))
+
+        return metadata
+
+    @classmethod
+    def _subtitle(cls, main_title, subtitled_title):
```
leonardr: This might be a useful thing to put in core/util, but it's okay here for now.

tests/__init__.py
```diff
@@ -1,9 +1,15 @@
...
 
-from ..core.testing import (
+from core.testing import (
     DatabaseTest,
     package_setup,
 )
 
 package_setup()
+
+def sample_data(filename, sample_data_dir):
```
leonardr: I like this change a lot.

leonardr:
It looks like I have a lot of comments on this and ideas for future work, but nothing that should actually prevent this from being committed as a good-enough-for-now solution except for bumping up the weight of the equivalency necessary to look up an ISBN associated with a non-ISBN identifier.

novelist.py
courte: Started looking at this and Overdrive-source ISBNs are returning effectively `null` results at about a 20% rate. I'm still trying to get a sense of why this is (possibly the library audiobook isbn format, possibly something else), and I'll make changes accordingly once I know -- or at least set something up so that we check back sooner rather than later.

courte:
A lot of changes since the original PR. Seeking the answer to "Why, Overdrive IDs, why??" led me to finish the full implementation of NoveList; I wanted to use the CoverageProvider to think through equivalencies.

Since the last comment, I've looked at a larger, more random set of Overdrive IDs and their equivalent ISBNs. Of the 250 identifiers run, more than 30% return no useful bibliographic data. There's no absolute format reason for works and their failure here, but audiobooks fail about 3x more than physical books.

I don't have a clear solution for this yet, but I'm hopeful that running all of the available ISBNs against NoveList with a central NoveList ID will help us get around the dependence on the licensing datasource's original ISBN. Right now, OCLC equivalencies can end up with really low strength -- not high enough to trust. But if multiple OCLC equivalents have the same NoveList identifier, and the data from that identifier matches OverDrive to whatever degree we decide, maybe we can come to trust those identifiers.

This is particularly relevant for incorporating NoveList features in circulation because the circ manager has drastically fewer equivalencies than the wrangler does. If we can establish trust on the wrangler, we can send circ managers an equivalent ISBN that we *know* will return NoveList results (and thus recommendations) in cases where the Overdrive-provided ISBN does not.

bin/novelist_coverage_provider
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env python
+"""Assign subjects to genres."""
```
leonardr: This docstring is wrong.

novelist.py
```diff
@@ -0,0 +1,309 @@
...
+        password = config.get(Configuration.NOVELIST_PASSWORD)
+        if not (profile and password):
+            raise ValueError("No NoveList client configured.")
+        return cls(_db, profile, password)
+
+    def __init__(self, _db, profile, password):
+        self._db = _db
+        self.profile = profile
+        self.password = password
+        self.source = DataSource.lookup(self._db, DataSource.NOVELIST)
```
leonardr: I recently decided that storing DataSources like this is bad practice because if this object ever gets used in a multithreaded web app, it'll break pretty quickly. I know that this object isn't currently used in a web app, but it might easily be in a metadata wrangler admin app. So I'd recommend (at least for now) doing what I did in core/nyt.py, something like this:

```
    @property
    def source(self):
        return DataSource.lookup(_db, DataSource.NOVELIST)
```

novelist.py
```diff
@@ -0,0 +1,309 @@
...
+            self.log.error("Identifiers without an ISBN equivalent can't \
+                    be looked up with NoveList: %r" % identifier)
+            return None
+
+        return [metadata for metadata in lookup_metadata if metadata]
+
+    def lookup(self, identifier):
+        """Requests NoveList metadata for a particular identifier
+
+        :return: None, a Metadata object, or a list of Metadata objects
```
leonardr: It would be easier on the caller's end if this reliably returned a (potentially empty) list of Metadat objects.

novelist.py
```diff
@@ -0,0 +1,309 @@
...
+            ClientIdentifier=client_identifier, ISBN=identifier.identifier,
+            version=self.version, profile=self.profile, password=self.password
+        )
+        url = self._build_query(params)
+        self.log.debug("NoveList lookup: %s", url)
+        representation, from_cache = Representation.cacheable_post(
+            self._db, unicode(url), params, max_age=self.MAX_REPRESENTATION_AGE
+        )
+
+        # Confirm that the representation was successful.
```
leonardr: Argh, this is really tricky. It would be better to check this information during `cacheable_post` in the first place, so that bad representations don't get put in the database at all.

You can do this by defining your own version of the `do_post` function defined be `cacheable_post`. Something like this:

```
        def do_post(url, headers, **kwargs):
            kwargs.update({'data' : params})
            response = cls.simple_http_post(url, headers, **kwargs)
            if response.status_code == 403:
                raise Exception("Invalid NoveList credentials")
            if response.content.startswith('"Missing'):
                raise Exception("Invalid Novelist parameters: %s" % error)
            return response.status_code, response.headers, response.content
```
        representation, from_cache = Representation.get(do_get=novelist_post, ...)
```

File: novelist.py [hunk 9]
```diff
@@ -0,0 +1,309 @@
...
+        book_info = lookup_info['TitleInfo']
+        if book_info:
+            novelist_identifier = book_info.get('ui')
+        if not book_info or not novelist_identifier:
+            # NoveList didn't know the ISBN. Delete the cache and return None.
+            client_identifier = lookup_info['ClientIdentifier']
+            cached = self._db.query(Representation).\
+                    filter(Representation.url.like(
+                        "%ClientIdentifier="+client_identifier+"%"
+                    ))
```
leonardr: I think it would be better to leave the cached document in place. We can always check again later and use max_age to see if NoveList has new information.

novelist.py
leonardr: Actually, looking further down the code I see you say that you expect to only get one piece of data here -- all ISBNs for a literary work will return the same information, and you throw an exception if the ISBNs turn out to represent more than one literary work.

So I would expect this to either stop processing ISBNs as soon as it finds some data, or to gather data for all the ISBNs and choose the most commonly occuring item (reducing the chance that a single bad ISBN ruins everything).

In either case this method would return only one piece of information.

novelist.py
```diff
@@ -0,0 +1,309 @@
...
+        self._db = _db
+        self.api = NoveListAPI.from_config(self._db)
+        self.output_source = DataSource.lookup(self._db, DataSource.NOVELIST)
+
+        super(NoveListCoverageProvider, self).__init__(
+            "NoveList Coverage Provider", [Identifier.ISBN],
+            self.output_source, workset_size=25
+        )
+
+    def process_item(self, identifier):
```
leonardr: This needs a test. You can test it by creating a dummy API object and feeding it Metadata objects to return when lookup() is called.

If it turns out that it's a good idea to make `lookup` return a single item, as I mentioned earlier, then this should become simple enough that a test is less important. I'd probably still test it though.

novelist.py
```diff
@@ -0,0 +1,309 @@
...
+
+        # If nothing interesting comes from the API, ignore it.
+        if not (metadata.measurements or metadata.series_position or
+                metadata.series or metadata.subjects or metadata.links or
+                metadata.subtitle):
+            return None
+
+        return metadata
+
+    def get_series_information(self, metadata, series_info, book_info):
```
leonardr: This needs some tests.

leonardr:
This is some good stuff. I have a lot of comments but it's coming along nicely.

You don't need to do this in this branch, but before you actually run all the ISBNs in the world through the NoveList API, we should come up with a way to store the appeal info and any other useful info. We don't have to do anything with it yet, but storing it now will probably save us from having to run through all the ISBNs a second time once we decide to start using it.

novelist.py
courte: Should I go ahead and do this (or plan to) in all the APIs and CoverageProviders in the wrangler?

leonardr: Yes please.

novelist.py
courte: There's no way to avoid some kind of Representation being made right now in `Representation.get`, but I can make one with the Exception captured. I've decided to pass a "reviewer" function from Novelist into `cacheable_post` instead of recreating `cacheable_post` wholesale.

novelist.py
```diff
@@ -94,6 +93,34 @@ def _confirm_same_identifier(self, metadata_objects):
         ])
         return len(novelist_ids)==1
 
+    def choose_best_metadata(self, metadata_objects, identifier):
```
leonardr: Usually in this situation I return a confidence score along with the data. (The code that calculates recursive equivalent identifiers is the closest thing in the Simplified code.) 

In this case you could return a 2-tuple (best_choice, confidence) and confidence would be the number of times the best choice showed up, divided by the total number of items.

It's up to you since your code is probably the only code that will call this method, but in situations like this I like having a rough picture of how reliable the data is.

leonardr:
:+1:  but with a comment about choose_best_metadata.

PR #67, opened 2016-06-10
Workset size to batch size

leonardr:
This branch updates core and changes the `workset_size` argument used in CoverageProvider constructors to the new name, `batch_size`.


aslagle:
:+1:

PR #68, opened 2016-06-15
Fix identifier resolution

leonardr:
The identifier resolution monitor has been broken for a while. We didn't notice because it has no test coverage. It has no test coverage because it runs through a bunch of CoverageProviders which themselves have no test coverage.

The purpose of this branch is to make the minimal changes necessary to get the identifier resolution script working again. I also add test coverage to one of the underlying CoverageProviders, and I reduce the amount of code that needs to be covered before we're confident that the identifier resolution script won't break again without us knowing.

### DIFF
--- a/content_server.py
+++ b/content_server.py
@@ -19,7 +19,9 @@ class ContentServerException(Exception):
     pass
 
 class ContentServerCoverageProvider(CoverageProvider):
-    """Checks the OA Content Server for Records"""
+    """Checks the OA Content Server for metadata about Gutenberg books
+    and books identified by URI.
+    """
 
     CONTENT_SERVER_RETURNED_WRONG_CONTENT_TYPE = "Content Server served unhandleable media type: %s"
 
@@ -57,7 +59,8 @@ class ContentServerCoverageProvider(CoverageProvider):
             return CoverageFailure(
                 identifier,
                 self.CONTENT_SERVER_RETURNED_WRONG_CONTENT_TYPE % (
-                    content_type)
+                    content_type),
+                data_source
             )
         
         editions, licensepools, works, messages = self.importer.import_from_feed(
--- a/content_server.py
+++ b/content_server.py
@@ -12,6 +12,7 @@ from core.coverage import (
     CoverageProvider,
     CoverageFailure,
 )
+from core.util.http import BadResponseException
 
 class ContentServerException(Exception):
     # Raised when the ContentServer can't connect or returns bad data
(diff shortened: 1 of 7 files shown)


content_server.py
```diff
@@ -40,10 +43,27 @@ def __init__(self, _db):
         )
 
     def process_item(self, identifier):
-        response = self.content_server.lookup([identifier])
-        self.check_response_for_errors(response)
-
-        editions, messages, next_links = self.importer.import_from_feed(
```
leonardr: The actual thing that broke here was the return signature of import_from_feed.

content_server.py
```diff
@@ -53,35 +73,11 @@ def process_item(self, identifier):
             if edition_identifier == identifier:
                 return identifier
         expect = identifier.urn
-        for message_identifier, status_message in messages.items():
```
leonardr: This code isn't necessary anymore because import_from_feed returns CoverageFailure objects if there's a problem.

monitor.py
```diff
@@ -104,58 +104,6 @@ def __init__(self, _db):
         self.image_scaler = ImageScaler(self._db, self.image_mirrors.values())
         self.oclc_linked_data = LinkedDataCoverageProvider(self._db)
 
-
```
leonardr: This is very messy code that's not needed very often. I moved it into a standalone script so that it's less urgent we provide test coverage.

monitor.py
```diff
@@ -206,7 +154,8 @@ def resolve_equivalent_oclc_identifiers(self, identifier):
     def resolve_viaf(self, work):
         """Get VIAF data on all contributors."""
         viaf = VIAFClient(self._db)
-        for edition in work.editions:
+        for pool in work.license_pools:
```
leonardr: This broke when we broke the direct connection between Work and Edition.

monitor.py
```diff
@@ -215,7 +164,8 @@ def resolve_viaf(self, work):
 
     def resolve_cover_image(self, work):
         """Make sure we have the cover for all editions."""
-        for edition in work.editions:
+        for pool in work.license_pools:
```
leonardr: Ditto.

scripts.py
```diff
@@ -23,14 +24,15 @@
...
     WorkProcessingScript,
     Script,
     RunMonitorScript,
 )
 from viaf import VIAFClient
 from core.util.permanent_work_id import WorkIDCalculator
 
-class RunIdentifierResolutionMonitor(RunMonitorScript):
+class RunIdentifierResolutionMonitor(RunMonitorScript, IdentifierInputScript):
     """Run the identifier resolution monitor.
```
leonardr: This script had its own custom way of parsing identifiers. I made it a real IdentifierInputScript.

scripts.py
```diff
@@ -377,3 +378,68 @@ def run(self, name):
         print "-" * 40
         print "CLIENT ID: %s" % collection.client_id
         print "CLIENT SECRET: %s" % plaintext_client_secret
+
+
+class AddMissingUnresolvedIdentifiersScript(Script):
```
leonardr: Here's the code I moved out of IdentifierResolutionMonitor.

tests/test_coverage.py
```diff
@@ -0,0 +1,113 @@
+from nose.tools import (
```
leonardr: I added test coverage of ContentServerCoverageProvider before I realized adding test coverage for everything used by IdentifierResolutionMonitor would be a huge job.

courte:
:+1:

PR #72, opened 2016-06-16
Remove NoveList Select code from metadata; it's in Circulation now.

courte:
Don't worry about this code. It's in a better place. (The Circulation Manager, to be precise.)


leonardr:
:+1: :truck:

PR #77, opened 2016-07-19
Fix tests to bring metadata up to date with latest core api.

leonardr:


tests/test_oclc.py
```diff
@@ -169,13 +169,12 @@ def test_extract_single_work(self):
         # Most of the contributors have LC and VIAF numbers, but two
         # (Cliffs Notes and Rockwell Kent) do not.
         eq_(
-            [None, None, u'n50025038', u'n50025038', u'n50050335',
-             u'n79006936', u'n79059764', u'n79059764', u'n79059764',
+            [None, None, u'n50025038', u'n50050335', u'n79006936',
```
leonardr: `.contributors` now returns a set of distinct contributors to avoid redundancy.

tests/test_scripts.py
```diff
@@ -106,12 +106,16 @@ def test_merge_contributors(self):
...
             )
             self._db.commit()
         eq_(0, len(self.edition1.contributors))
         self.script.merge_contributors(self.edition1.primary_identifier)
         eq_(2, len(self.edition1.contributors))
-        eq_("Denzel Washington", self.edition1.contributors[0].name)
+        eq_(
+            ["Denzel Washington"] * 2, 
+            [x.name for x in self.edition1.contributors]
+        )
```
leonardr: I don't understand why `merge_contributors` creates two distinct `Contributor` objects for Denzel Washington. I'm pretty sure that was the old behavior, and maybe it wasn't as noticeable back when `.contributors` was expected to return the same `Contributor` more than once, but it seems worrying.

courte: I can't work out why this is happening either - still spelunking - but I agree it shouldn't be the behavior.

courte: Sorry - it's because the test is poorly written. It's testing against an edition with an OCLC Work ID and an edition with an OCLC Work Number to make sure they both works, but it gives them each a different contributor with the same name.

courte:
:+1:

PR #80, opened 2016-07-29
If OCLC Linked Data identifies an author by VIAF ID and doesn't provide a name, quickly look up their VIAF ID to find that name.

leonardr:
Sometimes OCLC Linked Data identifies an author solely by VIAF and we can't write the data to the database because the metadata layer won't create a Contributor object with no sort_name. This leads to a lot of failures in the metadata wrangler's identifier resolution code. 

This is maybe not the right thing to do all the time, but it's very useful to guarantee that every person in the database has a name, so rather than remove this restriction I now avoid it by doing a quick VIAF lookup on the spot and filling in the rest of the information.

This branch also includes what I believe is the first end-to-end test of `LinkedDataCoverageProvider.process_item()`.

### DIFF
--- a/oclc.py
+++ b/oclc.py
@@ -38,6 +38,7 @@ from core.metadata_layer import (
 )
 from core.util import MetadataSimilarity
 
+from viaf import VIAFClient
 
 class OCLC(object):
     """Repository for OCLC-related constants."""
@@ -539,8 +540,8 @@ class OCLCLinkedData(object):
 
         self.log.info("Processing edition %s: %r", oclc_id, titles)
         metadata = Metadata(self.source)
-        metadata.primary_identifier, new = Identifier.for_foreign_id(
-            self._db, oclc_id_type, oclc_id
+        metadata.primary_identifier, new = IdentifierData(
+            type=oclc_id_type, identifier=oclc_id
         )
         if titles:
             metadata.title = titles[0]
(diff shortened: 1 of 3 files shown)


oclc.py
```diff
@@ -539,8 +540,8 @@ def book_info_to_metadata(self, subgraph, book_info):
 
         self.log.info("Processing edition %s: %r", oclc_id, titles)
         metadata = Metadata(self.source)
-        metadata.primary_identifier, new = Identifier.for_foreign_id(
-            self._db, oclc_id_type, oclc_id
+        metadata.primary_identifier, new = IdentifierData(
+            type=oclc_id_type, identifier=oclc_id
```
leonardr: Sticking an Identifier object here instead of an IdentifierData was a lazy hack which stopped working once I started seriously calling methods on this Metadata object.

oclc.py
```diff
@@ -1290,14 +1292,26 @@ def process_item(self, identifier):
             self.log.info("Processing identifier %r", identifier)
 
             for metadata in self.api.info_for(identifier):
-                oclc_editions = metadata.primary_identifier.primarily_identifies
```
leonardr: Because I got rid of the earlier hack, I can no longer refer to `metadata.primary_identifier` as an `Identifier`.

oclc.py
```diff
@@ -1333,7 +1347,10 @@ def process_item(self, identifier):
             else:
                 exception = "OCLC raised an error: %r" % e
                 transient = True
-            return CoverageFailure(identifier, exception, transient=transient)
+            return CoverageFailure(
+                identifier, exception, data_source=self.output_source,
```
leonardr: The absence of `data_source` here sometimes meant that `CoverageFailure` objects couldn't be converted to `CoverageRecord` objects.

courte:
:+1:

PR #81, opened 2016-08-01
Always make sure to pass an Identifier object, not an IdentifierData, into Identifier.equivalent_to.

leonardr:
This branch fixes a problem where `IdentifierData` objects were being passed into data model methods that expected to take data model `Identifier` objects. It also corrects the test, which was sidestepping the problem by associating `Identifier` objects with `Metadata` objects where it should have been using `IdentifierData` objects.


courte:
:+1: I don't see another path into this method either.

PR #82, opened 2016-08-05
Move in complex URNLookupController code

leonardr:
This branch creates a subclass of URNLookupController which contains all the complicated code about registering UnresolvedIdentifiers and turning ISBNs into OPDS entries, code that used to be in core.

This code stands to be drastically improved, but the first step is isolating it inside the metadata wrangler codebase.

The general rule is that the core controller can turn an Identifier into an OPDS entry if that Identifier can be associated with a presentation-ready work. The metadata controller can do that as well, but it can also turn an ISBN into an OPDS entry if that ISBN has the appropriate coverage records.

If neither of these things is true, but the Identifier is an ISBN, an Overdrive identifier, a Gutenberg ID, or a URI, the metadata wrangler creates an UnresolvedIdentifier record for the Identifier, which is removed if and when the appropriate coverage records are provided.

This brings us to the only actual change in functionality in this branch. Currently if you ask metadata to look up a 3M ID, it will say "I'm working on it" but it will never be able to do the work. After this branch, if you ask metadata to look up a 3M ID it will refuse, _unless_ for whatever reason that 3M ID is associated with a presentation-ready work. Then it will work as before. (This will happen with 3M IDs that were in NYPL's collection and processed by the metadata wrangler before we discovered that metadata couldn't handle 3M IDs in general.)

Since we no longer process 3M IDs directly, this branch deletes ThreeMCoverImageMirror and removes ThreeMBibliographicCoverageProvider from the list of providers used when resolving UnresolvedIdentifiers.


controller.py
```diff
@@ -112,3 +117,181 @@ def remove_items(self):
...
+        """
+        pool = identifier.licensed_through
+        if not pool:
+            return None
+        work = pool.work
+        if not work or not work.presentation_ready:
+            return None
+        return work
+    
+    def can_resolve_identifier(self, urn, identifier):
```
courte: You'll want to remove `urn` as a parameter.

controller.py
```diff
@@ -112,3 +117,181 @@ def remove_items(self):
         )
 
         return feed_response(removal_feed)
+
+
+class URNLookupController(CoreURNLookupController):
+
+    UNRESOLVABLE_IDENTIFIER = "I can't gather information about an identifier of this type."
```
courte: This constants already exists in `CoreURNLookupController` (albeit with a slightly different message).

courte: Oh, but it's not being used there.

courte:
👍

PR #83, opened 2016-08-05
Author matching

daryanypl:
What it has:
New and improved VIAF request parsing.

What it doesn't have yet:
Marking equivalent contributors.
Maybe a couple extra unit tests are in order for the new sorting methods.


tests/test_viaf.py
```diff
@@ -20,76 +32,73 @@ def test_entry_with_wikipedia_name(self):
...
         xml = self.sample_data("mark_twain_no_wikipedia.xml")
 
         # The author is better known as Mark Twain, so this
         # name wins by popularity if we don't specify a name going in.
-        viaf, display, family, sort, wikipedia = self.parser.parse(xml, None)
-        eq_("50566653", viaf)
-        eq_("Mark Twain", display)
-        eq_("Twain", family)
-        eq_(None, wikipedia)
+        #viaf, display, family, sort, wikipedia = self.parser.parse(xml, None)
```
leonardr: Take this line out.

tests/test_viaf.py
```diff
@@ -20,76 +32,73 @@ def test_entry_with_wikipedia_name(self):
...
         # Even if we go in expecting something like "Sam Clemens",
         # we get the consensus result.
-        viaf, display, family, sort, wikipedia = self.parser.parse(
-            xml, working_display_name="Samuel Langhorne Clemens")
-        eq_("50566653", viaf)
-        eq_("Mark Twain", display)
-        eq_("Twain, Mark", sort)
-        eq_("Twain", family)
-        eq_(None, wikipedia)
+        #viaf, display, family, sort, wikipedia = self.parser.parse(
```
leonardr: Same here.

viaf.py
```diff
@@ -2,60 +2,271 @@
...
+
+
+
 
 class VIAFParser(XMLParser):
 
     NAMESPACES = {'ns2' : "http://viaf.org/viaf/terms#"}
 
     log = logging.getLogger("VIAF Parser")
+    log.setLevel(logging.INFO)  # TODO: remove
```
leonardr: Take this out.

viaf.py
```diff
@@ -2,60 +2,271 @@
...
+        """
+        if not given and not family:
+            return None
+        if family and not given:
+            display_name = family
+        elif given and not family:
+            display_name = given
+        else:
+            display_name = given + ' ' + family
+        if extra and not extra.startswith('pseud'):
```
leonardr: There are a number of others. I've seen "Leo Tolstoy, graf", though I'm not sure what that means. But this is okay for now.

viaf.py
```diff
@@ -2,60 +2,271 @@
...
+        name2 = cls.prepare_contributor_name_for_matching(name2)
+        match_ratio = fuzz.ratio(name1, name2)
+        return match_ratio
+
 
     @classmethod
     def name_matches(cls, n1, n2):
+        """ Returns true if n1 and n2 are identical strings.  
+        Ignores case, periods, commas, spaces after single-letter initials.
+        """
```
leonardr: It doesn't really do half of these things; change the docstring and add the code later?

viaf.py
```diff
@@ -2,60 +2,271 @@
...
+                            match_confidences["title"] = 90
+                            match_confidences["total"] += 0.8 * match_confidences["title"]
+                            # match is good enough, we can stop
+                            break
+
+                        '''
+                        Fixes issue where 
+                        <ns1:title>Britain, detente and changing east-west relations</ns1:title> (with accented e in detente)
+                        doesn't match "Britain, Detente and Changing East-West Relations" in our DB.
+                        '''
```
leonardr: You can handle this specific problem by removing non-alphanumerics from the string (as permanent work ID does) and comparing that.

leonardr:
Equivalent contributors can wait. 

I would like to see some more unit tests illustrating some of the most common/interesting cases, such as the fact that a good match higher in the list (in terms of library popularity) penalizes a match lower in the list, but a bad match high in the list doesn't penalize matches below it.

Since Courteney has overall responsibility for the metadata wrangler, I'd like you to go through the code with her before merging. Not necessarily the diff. I wrote the original code, so the diff is useful for me, but I don't think Courteney has spent a lot of time with the original implementation, so going through the new implementation as though the old one didn't exist would probably be more useful.

tests/test_viaf.py
```diff
@@ -98,35 +107,78 @@ def test_ignore_results_if_author_not_in_viaf(self):
...
         xml = self.sample_data("lancelyn_green.xml")
         name = "Green, Roger Lancelyn"
-        contributor, new = self._contributor(name)
-        viaf, display_name, family_name, sort_name, wikipedia_name = self.parser.parse_multiple(xml, working_sort_name=name)
-        eq_("29620265", viaf)
-        eq_("Roger Lancelyn Green", display_name)
-        eq_("Green", family_name)
-        eq_("Roger_Lancelyn_Green", wikipedia_name)
+        #contributor, new = self._contributor(name)
+        #viaf, display_name, family_name, sort_name, wikipedia_name = self.parser.parse_multiple(xml, working_sort_name=name)
```
courte: Also here.

tests/test_viaf.py
```diff
@@ -98,35 +107,78 @@ def test_ignore_results_if_author_not_in_viaf(self):
...
         xml = self.sample_data("kate_lister.xml")
         name = "Lister, Kate"
-        viaf, display_name, family_name, sort_name, wikipedia_name = self.parser.parse(
-            xml)
-        eq_("68169992", viaf)
-        eq_(None, display_name)
-        eq_(None, family_name)
-        eq_("Lister, Kate", sort_name)
-        eq_(None, wikipedia_name)
+        #viaf, display_name, family_name, sort_name, wikipedia_name = self.parser.parse(xml)
```
courte: Ditto.

viaf.py
```diff
@@ -2,60 +2,271 @@
...
+                # the sort name matches perfectly and the title isn't there to contradict
+                # return with highest score to assert this contributor as The correct candidate.
+                match_confidences["total"] += match_confidences["sort_name"]
+                report_string += ", sort_name match and no title, return 9999 (%s)" % match_confidences["sort_name"]
+                report_string += ", mc[total]= %s" % match_confidences["total"]
+                cls.log.debug(
+                    "weigh_contributor: " + report_string
+                )
+                return match_confidences["total"]
+            '''
```
courte: yr code @ you: delete me, please!

leonardr:
I think this is ready to merge once you remove the commented-out code that Courteney and I pointed out. I'd still like to see that unit test I mentioned, and now is the best time to write the test while the code is fresh in your mind.

tests/test_viaf.py
daryanypl: done

tests/test_viaf.py
daryanypl: done

tests/test_viaf.py
daryanypl: done

tests/test_viaf.py
daryanypl: done

viaf.py
daryanypl: done

viaf.py
daryanypl: (Graf means Count, dude was one of those snooty aristocrats who slummed it in the villages and thought themselves spiritual and just.  Not that I'm bitter.)

viaf.py
daryanypl: Done.  Not really using that method anymore.

viaf.py
daryanypl: *cry*

viaf.py
daryanypl: But what about foreign language titles?  I think we need some sort of package that would find common equivalencies between foreign letters and ascii letters.

leonardr: Check out normalize_title in permanent_work_id.py. The Unicode title is normalized to NFKD, which turns [E WITH ACUTE] into [E] and [ACUTE], and then the non-ascii (the ACUTE) is removed.

leonardr: And, importantly, if there are _no_ ASCII characters in the title (because it's in a different script) the whole idea of trying to get an ASCII version is junked.

daryanypl: OK, traced the code.  Looks like conversion to nfkd is performed in one of the methods called by title_match_ratio.

leonardr:
:+1:

PR #84, opened 2016-08-15
Very brief start to replacing UnresolvedIdentifier with CoverageRecord.

daryanypl:


courte:
This PR:
  - Gives incoming identifiers a CoverageRecord with a transient failure.
  - Turns the `IdentifierResolutionMonitor` into an `IdentifierResolutionCoverageProvider` that works as expected.
  - Adds a migration to change existing UnresolvedIdentifiers into CoverageRecords.
  - Gets existing tests (pulled from server core) working.

This PR does NOT:
  - Determine the order in which identifiers receive coverage, which was an issue in the Monitor days when time would be wasted trying to cover identifiers that repeatedly raised errors.
  - Fix any existing errors in the finalization process for a work.

migration/20160830-change-unresolved-identifiers-into-coverage-records.py
```diff
@@ -0,0 +1,54 @@
...
+        if not record.exception:
+            # The UnresolvedIdentifier didn't have an exception, so this
+            # identifier hadn't been run at all yet.
+            # We'll give it the default lookup message.
+            record.exception = URNLookupController.NO_WORK_DONE_EXCEPTION
+
+        print "Resolution Coverage Record created for %r" % identifier
+    else:
+        print "Resolution Coverage Record already exists for %r" % identifier
+_db.commit()
```
leonardr: Should we delete the UnresolvedIdentifier here, or are we doing that later, maybe when we drop the table?

courte: I think when we drop the table, so we can make sure that this has run successfully.

As a check, it could be good to add some logging about how many coverage records exist before and after the script is done, to ensure none are missed. (Though if they're missed, the CoverageProvider will still nab 'em.)

leonardr:
:+1:

PR #85, opened 2016-08-19
Update to use the new OPDSMessage interface

leonardr:
This branch changes the metadata wrangler's CollectionController to use the new OPDSMessage class (see https://github.com/NYPL-Simplified/server_core/pull/376) to convey error messages, rather than the old system based on fake `<entry>` tags.


courte:
The tests are failing at the moment due to an import issue. But I like the new message format, very sleak, very lean. 💪

courte:
👍

PR #88, opened 2016-09-02
Fix LinkedDataCoverageProvider error with VIAF

### LINKED ISSUE #87: OCLCLinkedDataCoverageProvider fails during identifier resolution process
VIAF's `lookup_by_viaf` method now returns `contributor_data, match_confidences, contributor_titles` instead of the five values expected by the OCLCLinkedDataCoverageProvider in the error below.

```
2016-08-31 05:58:50,959:Identifier Resolution Manager:ERROR:monitor.py:FAILURE on Overdrive ID/0ea9b628-da1d-4125-8f75-b5aa2e2556f8 ID=4826992 prim_ed=302836 ("Identity Unknown"): Traceback (most recent call last):
  File "/home/ec2-user/metadata/core/monitor.py", line 138, in resolve_and_handle_result
    self.resolve(unresolved_identifier)
  File "/home/ec2-user/metadata/core/monitor.py", line 174, in resolve
    self.finalize(unresolved_identifier)
  File "/home/ec2-user/metadata/monitor.py", line 103, in finalize
    self.resolve_equivalent_oclc_identifiers(identifier)
  File "/home/ec2-user/metadata/monitor.py", line 146, in resolve_equivalent_oclc_identifiers
    self.oclc_linked_data.ensure_coverage(oclc_id)
  File "/home/ec2-user/metadata/core/coverage.py", line 436, in ensure_coverage
    [identifier]
  File "/home/ec2-user/metadata/core/coverage.py", line 235, in process_batch_and_handle_results
    results = self.process_batch(batch)
  File "/home/ec2-user/metadata/core/coverage.py", line 307, in process_batch
    result = self.process_item(item)
  File "/home/ec2-user/metadata/oclc.py", line 1314, in process_item
    i.viaf, i.display_name, i.family_name, i.sort_name, i.wikipedia_name = r
ValueError: need more than 3 values to unpack
```

Update the coverage provider to work within its means.

courte:
The main changes here were:
 1. Changing both VIAF and the LInkedDataCoverageProvider to ensure the `VIAFClient#lookup_by_viaf` response returns what's expected. Adjusted the provider's tests accordingly, too.
 2. `VIAFClient#lookup_by_name` now returns a single best candidate instead of a list, which allowed me to fix a couple broken references to that method as well.
 3. Refactoring the Canonicalizer so it could be passed into `Metadata#apply`. It's also now in a Controller and out of app.py, which feels like a win-win. 😄 

Musings:
- As you can see in the test, VIAF is now being hit _twice_ for each candidate found during the OCLC LinkedData coverage process. And -- since VIAF isn't holding on to Representations -- it is actually making the API call both times. This is going to extend the time needed to run the provider. 
- I had a thought to create a VIAFCandidate object (or something like it) instead of passing around a tuple with extraneous bits, but I didn't think it was worth the time or effort at this point.

Fixes #87.

### DIFF
--- a/canonicalize.py
+++ b/canonicalize.py
@@ -102,11 +102,11 @@ class AuthorNameCanonicalizer(object):
         self.log.debug("Attempting to canonicalize %s", display_name)
         contributors = self._db.query(Contributor).filter(
             Contributor.display_name==display_name).filter(
-                Contributor.name != None).all()
+                Contributor.sort_name != None).all()
         sort_name = None
         if contributors and False:
             # Yes, awesome. Use this name.
-            sort_name = contributors[0].name
+            sort_name = contributors[0].sort_name
             self.log.debug(
                 "Found existing contributor for %s: %s",
                 display_name, sort_name
--- a/canonicalize.py
+++ b/canonicalize.py
@@ -1,18 +1,21 @@
 """Use external services to canonicalize names."""
-from nose.tools import set_trace
 import logging
-from oclc import OCLCLinkedData
+import re
+from nose.tools import set_trace
+
+from core.model import (
+    Contributor,
+    Identifier,
+)
 from core.util import MetadataSimilarity
 from core.util.personal_names import (
     display_name_to_sort_name,
     is_corporate_name,
 )
+
+from oclc import OCLCLinkedData
 from viaf import VIAFClient
-from core.model import (
-    Contributor,
-    Identifier,
-)
-import re
+
 
 class CanonicalizationError(Exception):
     pass
@@ -53,7 +56,7 @@ class AuthorNameCanonicalizer(object):
 
         return author_name
 
-    def canonicalize(self, identifier, display_name):
+    def canonicalize_author_name(self, identifier, display_name):
         """Canonicalize a book's primary author given an identifier and a
         display name.
 
@@ -71,7 +74,8 @@ class AuthorNameCanonicalizer(object):
             identifier = None
         if not identifier and not display_name:
             raise CanonicalizationError(
-                "Neither useful identifier nor display name was provided.")
+                "Neither useful identifier nor display name was provided."
+            )
 
         # From an author name that potentially names multiple people,
         # extract only the first name.
@@ -84,8 +88,9 @@ class AuthorNameCanonicalizer(object):
             if v:
                 return v
 
-        # All our techniques have failed. Woe!
-        return None
+        # All our techniques have failed. Woe! Let's just try to finagle
+        # this provided display name into a sort name.
+        return self.default_name(display_name)
 
     def default_name(self, display_name):
         shortened_name = self.primary_author_name(display_name)
--- a/canonicalize.py
+++ b/canonicalize.py
@@ -125,10 +125,10 @@ class AuthorNameCanonicalizer(object):
                 m = self.VIAF_ID.search(uri)
                 if m:
                     viaf_id = m.groups()[0]
-                    viaf_id, display_name, family_name, sort_name, wikipedia_name = (
-                        self.viaf.lookup_by_viaf(
-                            viaf_id, working_display_name=display_name))
-                    if sort_name:
+                    contributor_data = self.viaf.lookup_by_viaf(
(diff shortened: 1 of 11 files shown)


canonicalize.py
```diff
@@ -53,7 +56,7 @@ def primary_author_name(self, author_name):
 
         return author_name
 
-    def canonicalize(self, identifier, display_name):
+    def canonicalize_author_name(self, identifier, display_name):
```
courte: This is the method name that Metadata needs to be able to call.

oclc.py
```diff
@@ -1353,6 +1351,32 @@ def process_item(self, identifier):
...
+        This is particularly crucial for contributors identified solely
+        by VIAF IDs (and no sort_name), as it raises errors later in the
+        process.
+        """
+        for contributor_data in metadata.contributors:
+            viaf_contributor_data = self.viaf.lookup_by_viaf(
+                contributor_data.viaf,
+                working_sort_name=contributor_data.sort_name,
+                working_display_name=contributor_data.display_name
+            )[0]
```
leonardr: Looking at the VIAF code I think it's quite possible lookup_by_viaf could return an empty list, which would cause a crash here.

viaf.py
```diff
@@ -136,7 +136,7 @@ def weigh_contributor(cls, candidate, working_sort_name, known_titles=None, stri
         # the given author, make sure that one of the working names
         # shows up in a name record.
         if strict:
-            if len(match_confidences) == 0:
+            if not match_confidences:
```
leonardr: Why this change? Usually this type of change goes in the opposite direction (from a generic boolean comparison to a more specific check). Is match_confidences liable to be None?

viaf.py
```diff
@@ -369,28 +369,39 @@ def order_candidates(self, contributor_candidates, working_sort_name,
         :return: the list of tuples, ordered by percent match, in descending order 
         (top match first).
         """
-
-        if len(contributor_candidates) == 0:
+        if not contributor_candidates:
```
leonardr: Similar question here.

viaf.py
```diff
@@ -751,39 +783,42 @@ def lookup_by_viaf(self, viaf, working_sort_name=None,
...
                 break
 
             contributor_candidates.extend(candidates)
             page += 1
 
-        return contributor_candidates
-
-
-
+        return self.select_best_match(contributor_candidates, sort_name)
```
leonardr: I bet Darya's opinion on this will be a lot stronger than mine, but what's the advantage of calling select_best_match here (and changing the contract of the method) instead of calling it on line 201, the code that calls the method?

leonardr:
I think you could get rid of that second request without having to track everything down by passing a `max_age` of... six months or so into `Representation.get`.

viaf.py
courte: There are a few of places outside of VIAF that call `lookup_by_name` and expect a single ContributorData (or rather, the former 5-tuple list of viaf, sort_name, wikipedia etc. etc.). I don't want those locations to have to know about selecting a best match from a list.

At first I changed it to a variable passed in to determine whether we select the best match or just return a list, but then realized that even the place where we use lookup_by_name in the VIAFClient just wanted the single best result. There's not really a place yet in the wrangler where we call lookup_by_name and want to sift through the results.

viaf.py
courte: It's a knee-jerk to past assumptions of list-hood. I'll change it back. match_confidences is always a dict.

viaf.py
leonardr: OK, if it's a general change I think it's the right thing to do. You might want some way of getting all the names, and have `lookup_by_name` call that other method and wrapping it in `select_best_match`, but I'd wait until there's some specific case that needs that (the 'specific case' might be the unit tests).

courte:
I agree, I'll add that here shortly.

viaf.py
leonardr: Now that I know what's going on, I think your change is fine. Without knowing what's going on this sort of change always sets off alarm bells.

daryanypl:
👍

PR #89, opened 2016-09-02
Fix Identifier Resolution utility command

courte:
The script in `/bin` was broken after we changed the IdentifierResolutionMonitor to an IdentifierResolutionCoverageProvider. No longer.


leonardr:
:+1:

PR #90, opened 2016-09-13
Give ISBNs license pools

courte:
This branch:
- Stops using the core URNLookupController (😦 )
- Gives all identifiers "fake" internal license pools, intended entirely for processing purposes.
- Allows DataSource.INTERNAL_PROCESSING to offer licenses for just such purposes.
- Adjusts tests accordingly.
- Removes all 3M/Bibliotheca identifiers from the identifier resolution queue.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -172,7 +172,8 @@ class IdentifierResolutionCoverageProvider(CoverageProvider):
                 return self.transform_exception_into_failure(e, identifier)
 
         try:
-            self.finalize(identifier)
+            self.resolve_equivalent_oclc_identifiers(identifier)
+            self.process_work(identifier)
         except Exception as e:
             return self.transform_exception_into_failure(e, identifier)
 
--- a/coverage.py
+++ b/coverage.py
@@ -59,6 +59,7 @@ class IdentifierResolutionCoverageProvider(CoverageProvider):
     For ISBNs, make a bunch of Resources, rather than LicensePooled Editions.
     """
 
+    CAN_CREATE_LICENSE_POOLS = True
     LICENSE_SOURCE_NOT_ACCESSIBLE = (
         "Could not access underlying license source over the network.")
     UNKNOWN_FAILURE = "Unknown failure."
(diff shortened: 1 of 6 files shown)


migration/20160913-2-change-all-license-pools-sources-to-internal-processing.sql
```diff
@@ -0,0 +1,3 @@
+UPDATE licensepools SET data_source_id = (
+    SELECT id FROM datasources WHERE name = 'Library Simplified Internal Process'
+) WHERE id IS NOT NULL;
```
courte: This will turn all licensepools on the metadata wrangler into Internal Process license pools, thus divorcing the Metadata Wranglers license pools from any semblance of licensing. This gets rid of information about where the existing licensepool came from, but the existing license pool has no relevant meaning, so perhaps it's okay?

coverage.py
```diff
@@ -132,6 +133,13 @@ def process_item(self, identifier):
         """
         self.log.info("Ensuring coverage for %r", identifier)
 
+        license_pool = self.license_pool(identifier)
```
courte: LicensePool or bust.

tests/test_coverage.py
```diff
@@ -209,10 +221,15 @@ def test_process_item_succeeds_when_optional_provider_fails(self):
...
-        # Even though the coverage provider failed and an appropriate
-        # coverage record was created to mark the failure.
-        r = get_one(self._db, CoverageRecord, identifier=self.identifier)
-        eq_("What did you expect?", r.exception)
 
+        # An appropriate coverage record was created to mark the failure.
+        presentation_edition = DataSource.lookup(
+            self._db, DataSource.PRESENTATION_EDITION
+        )
+        r = self._db.query(CoverageRecord).filter(
```
courte: Because we're attempting to create a work in process_item now, some additional DataSource.PRESENTATION_EDITION coverage records are created.

migration/20160913-3-remove-unresolved-3m-identifiers-from-queue.sql
```diff
@@ -0,0 +1,5 @@
+delete from coveragerecords
+    where operation = 'resolve-identifier' and
+    identifier_id in (
+        select id from identifiers where type in ('3M ID', 'Bibliotheca ID')
+    );
```
courte: There are 17,875 UnresolvedIdentifiers with exceptions connected to 3M IDs. In the migration created on 20160830, they'll be tranformed into CoverageRecords that will never be resolved because the IdentifierResolutionCoverageProvider does not cover 3M IDs.

leonardr:
Wow, we were using the core URNLookupController instead of the subclass? :blush:

leonardr:
:+1:

PR #91, opened 2016-09-15
Fix VIAFClient#process_contributor

courte:
Minor fixes to typos that were raising errors during identifier resolution.

### DIFF
--- a/tests/test_viaf.py
+++ b/tests/test_viaf.py
@@ -238,34 +238,57 @@ class TestVIAFClient(DatabaseTest):
         contributor = self._contributor()[0]
 
         # If lookup returns an empty array (as in the case of
-        # VIAFParser#parse_multiple), #process_contributor returns None.
+        # VIAFParser#parse_multiple), the contributor is not updated.
         client.queue_lookup([])
-        eq_(client.process_contributor(contributor), None)
+        client.process_contributor(contributor)
+        eq_(contributor.sort_name, '2001')
+        eq_(contributor.display_name, None)
 
         def queue_lookup_result():
-            h = self.queue_file_in_mock_http("mindy_kaling.xml")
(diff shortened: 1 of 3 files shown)


tests/test_viaf.py
```diff
@@ -204,35 +207,78 @@ def test_birthdates(self):
...
+        # VIAFParser#parse_multiple), #process_contributor returns None.
+        client.queue_lookup([])
+        eq_(client.process_contributor(contributor), None)
+
+        def queue_lookup_result():
+            h = self.queue_file_in_mock_http("mindy_kaling.xml")
+            lookup = self.client.lookup_by_viaf(viaf="9581122", do_get=h.do_get)
+            client.queue_lookup(lookup)
+
+        # When lookup returns a candidate, the candidate is returned.
```
courte: Rephrase.

viaf.py
```diff
@@ -717,15 +716,15 @@ def process_contributor(self, contributor):
             # No good match was identified.
             return None
 
-        (selected_candidate, match_confidences, contributor_titles) = best_contributor_candidate
+        (selected_candidate, match_confidences, contributor_titles) = contributor_candidate
         # Is there already another contributor with this VIAF?
         if selected_candidate.viaf is not None:
```
courte: Update #process_contributor to import VIAF information on the relevant contributors, according to its use in IdentifierResolutionCoverageProvider#resolve.viaf. (Update RedoOCLCForThreeMScript#do_run.)

daryanypl:
I like it.  In the future, we might want to check the author's birth and death dates as well as display name when guarding against untoward merges.

👍

courte:
Ah, good idea for when they're available. With the number of duplicates in the database right now, we probably need to give more thought to how to handle and merge existing contributor duplicates overall. It warrants its own migration / script.

PR #95, opened 2016-09-21
Remove Bibliotheca/3M OCLC Classify coverage failures

courte:
3M Identifiers are no longer getting coverage from the OCLC Classify. This migration will remove any transient failure coverage records from the database.


daryanypl:
👍

Issue #66, opened 2016-06-09
Do more work when asked to get data for an ISBN


leonardr:
Up to this point we've been doing a minimal amount of work when asked to look up an ISBN. We find/scale/mirror a cover image, and that's about it. 

This was okay because the only time the metadata wrangler was actually asked to fill in data for an ISBN was when we were getting coverage for an Axis 360 book. Axis 360 has good metadata generally so it wasn't a big deal.

Unfortunately, I have discovered that 3M, like Axis 360, refuses to send bibliographic data for a book not in your collection. This means that the metadata wrangler can't really look up 3M IDs, because it can only get information about IDs that are in NYPL's collection. And 3M's metadata is minimal, so it's really important that we supplement it (if possible) with metadata from OCLC Linked DAta.

So, If the metadata wrangler is going to be useful for the 3M books in the collections of BPL and other libraries, we need to be able to run an ISBN through OCLC Linked Data and any other sources that take an ISBN. Then we need to create a Work for the ISBN, and return the consolidated data in response to a lookup.

leonardr:
If the ISBN is associated with a title and author upon initial registration, metadata can do title/author lookups in other APIs like OCLC Classify and Overdrive.

PR #100, opened 2016-10-20
Give works to ISBNs through OCLC Linked Data

courte:
This is a strategy for turning ISBNs into works by getting their title & author from OCLC. This PR moves to broader author extraction from OCLC Linked Data, which creates more data for the OCLC editions so that one of them can be used as a presentation edition.

**This is basically the opposite of our strategy with OCLC thus far** -- we're getting whatever data we can scrape together, including getting into evaluating author name quality, instead of trying to keep all of that information in its own little corner.

I think it needs:
- a better method for evaluating which edition from OCLC is the "best" or most accurate
- more nuanced author name evaluation

I also have a concern that if works created using this method are passed back into Circulation Managers, this data (with its Metadata Wrangler DataSource) could possibly be given priority over the data that Bibliotheca provides. I'm not sure where editions from the DataSource get placed in the spectrum.

I'm in the process of running this against 200 Bibliotheca-equivalent ISBNs to see what percentage get a work through this resolution process. On a much smaller sample size, I got 45% coverage.

A start to NYPL-Simplified/metadata_wrangler#66.


coverage.py
```diff
@@ -196,13 +199,46 @@ def finalize(self, identifier):
 
         self.resolve_equivalent_oclc_identifiers(identifier)
         if identifier.type==Identifier.ISBN:
+            self.generate_edition(identifier)
             # Currently we don't try to create Works for ISBNs,
             # we just make sure all the Resources associated with the
             # ISBN are properly handled. At this point, that has
             # completed successfully, so do nothing.
```
leonardr: Is this comment still accurate?

tests/test_coverage.py
```diff
@@ -161,6 +161,18 @@ def test_process_item_creates_license_pool(self):
...
+        # Prepare an identifier that already has a LicensePool through
+        # another source.
+        licensed = self._identifier(identifier_type=Identifier.OVERDRIVE_ID)
+        other_source = DataSource.lookup(self._db, DataSource.OVERDRIVE)
+        lp = LicensePool.for_foreign_id(
+            self._db, other_source, licensed.type, licensed.identifier
+        )[0]
+
+        self.coverage_provider.process_item(licensed)
+        eq_(lp, licensed.licensed_through)
```
leonardr: I don't understand what this is testing. Are you verifying that `licensed_through` didn't get overwritten by some other value?

tests/test_oclc.py
```diff
@@ -245,37 +268,43 @@ def test_all_authors_get_viaf_lookup(self):
...
+        # and not those who do not.
         lookup1 = (ContributorData(
                   viaf="1", display_name="Display Name",
                   family_name="Family", sort_name="Name, Sort",
                   wikipedia_name="Wikipedia_Name"), None, None)
         lookup2 = (ContributorData(
                    viaf="2", wikipedia_name="Robert_Jordan_(Author)",
                    biography="That guy."), None, None)
-        viaf.queue_lookup(lookup1, lookup2)
+        viaf.queue_lookup(lookup1, lookup2, "Unrequested lookup")
```
leonardr: After you finish this test do you need to verify that 'unrequested lookup' is still stuck in the mock VIAF queue?

courte:
When I ran 200 ISBNs, 124 of them ended up with a Work. Quality of title and author varies. It's not terrible, but getting the information directly from the source the way we discussed would be 💯.

tests/test_oclc.py
courte: I verify a few lines below (unless you're saying it's not necessary to do so?). In the latest commit I moved that verification and added a bit more of a comment to clarify.

leonardr: ...don't know what I was thinking, you do verify it immediately afterwards. Never mind!

courte:
@leonardr would you take another look at this next week? It still offers ~52% coverage (I did two more test runs, and I got 86/200 and 104/200 for two random batches of Bibliotheca-sourced ISBN equivalents).

When a Work isn't created, it's because the title and author couldn't be sourced from OCLC. Covers, descriptions, and summaries are still scraped together and returned as an entry.

PR #106, opened 2016-12-12
Correctly handle the difference between metadata wrangler's INTERNAL_PROCESSING and other 

leonardr:


content_cafe.py
```diff
@@ -6,6 +6,9 @@
 from bs4 import BeautifulSoup
 from suds.client import Client as SudsClient
 
+# Tone down the verbose Suds logging.
+logging.getLogger('suds').setLevel(logging.ERROR)
+
```
leonardr: Suds was driving me crazy.

coverage.py
```diff
@@ -68,8 +68,12 @@ def __init__(self, _db, batch_size=10, cutoff_time=None,
                  uploader=None, providers=None, **kwargs):
         output_source, made_new = get_one_or_create(
             _db, DataSource,
-            name=DataSource.INTERNAL_PROCESSING, offers_licenses=True,
+            name=DataSource.INTERNAL_PROCESSING
```
leonardr: By default, `offers_licenses` is set to False when this DataSource is created in core. This was attempting to create a second INTERNAL_PROCESSING data source, which caused a database integrity error. Since we always want offers_licenses to be True, I just changed it to set `offers_licenses` after fetching the `DataSource` object.

content_cafe.py
courte: 🙏 👌

PR #108, opened 2017-02-10
Sort name cleaning

daryanypl:


canonicalize.py
```diff
@@ -68,10 +71,10 @@ def canonicalize_author_name(self, identifier, display_name):
         not "DH Lawrence".  NYT commonly formats names like "DH
         Lawrence".
         """
-        if identifier and identifier.type != Identifier.ISBN:
+        #if identifier and identifier.type != Identifier.ISBN:
```
leonardr: Take this whole thing out since you've improved the use of identifiers.

canonicalize.py
```diff
@@ -139,26 +172,31 @@ def _canonicalize(self, identifier, display_name):
...
             self, identifier, display_name):
         """Try to find an author sort name for this book from
         OCLC Linked Data.
+
+        :param identifier: Must be of Identifier.ISBN type.
         """
         def comparable_name(s):
             return s.replace(",", "").replace(".", "")
 
+        set_trace()
```
leonardr: x

canonicalize.py
```diff
@@ -195,12 +233,22 @@ def comparable_name(s):
...
+        :param known_titles: A list of titles we know this author wrote 
+            (helps better match the VIAF results if there's more than one matching VIAF author record).
+        :return: Author name in Last, First format.
+        """
+        sort_name = None
         viaf_contributor = self.viaf.lookup_by_name(
-            None, display_name, best_match=True
+            sort_name=None, display_name=display_name, known_titles=known_titles
         )
+        set_trace()
```
leonardr: x

viaf.py
```diff
@@ -709,8 +721,15 @@ def process_contributor(self, contributor):
                 contributor.viaf, contributor.sort_name, contributor.display_name
             )
         else:
+            known_titles = {}
```
leonardr: I feel like this should be a set, rather than a dictionary whose keys are set to 1 (which is basically a set that's harder to read).

leonardr:
The known_titles thing is very smart. My main note here is that you don't have any code to handle the REPAIR_SORT_NAME_OPERATION you define in core. I think you're working on that code now, but as I say in my core review, I don't think you should go in that direction. If you see that a sort name is wrong you should do a VIAF lookup then and there.

canonicalize.py
daryanypl: Done.

viaf.py
daryanypl: done

scripts.py
```diff
@@ -38,6 +55,310 @@ def run(self):
...
+        while contributors:
+            my_query = query.offset(offset).limit(batch_size)
+
+            #print "query=%s" % dump_query(my_query)
+            contributors = my_query.all()
+
+            for contributor in contributors:
+                self.process_contributor(contributor)
+            offset += batch_size
+            #self._db.commit()
```
leonardr: I think if you're implementing script.run() you want to uncomment these `commit()` calls--otherwise nothing will be written to the database.

scripts.py
```diff
@@ -38,6 +55,310 @@ def run(self):
...
+        parsed_args = self.parse_command_line(_db=self._db, cmd_args=cmd_args)
+        self.mock_mode = parsed_args.mock
+
+        if self.mock_mode:
+            self.log.debug(
+                "This is mocked run, with metadata coming from test files, rather than live OneClick connection."
+            )
+            self.base_path = os.path.split(__file__)[0]
+            self.base_path = os.path.join(self.base_path, "tests")
+            self.canonicalizer = MockAuthorNameCanonicalizer(self._db)
```
leonardr: I mentioned changing the way you mock scripts and this is what I'm talking about. Rather than passing in a `mock_mode` variable I recommend passing in the AuthorNameCanonicalizer class you want to use.

scripts.py
```diff
@@ -220,6 +541,7 @@ def fix_identifier_with_equivalents(self, primary_identifier, equivalent_ids):
 
         equivalent_ids = primary_identifier.equivalent_identifier_ids(
             levels=6, threshold=0)
+        # TODO: equivalencies is not used?
```
leonardr: I think originally there was a nested loop, where instead of just looking at `primary_identifier` we looked at every identifier in `equivalent_ids`. You can take this out if you like.

scripts.py
daryanypl: That code line is confusing -- the script is meant to scroll through the database without making any changes.  I added some docs.

scripts.py
daryanypl: Using mock_mode allows me to run the script with controlled input and without consequence not just from a unit test, but also from command line, and using a production database.  I'd like to understand these OneClick collections, and the difficulty they posed.

leonardr:
:+1:

PR #109, opened 2017-02-23
Rename Collection to Catalog

leonardr:
This branch makes sure my change to https://github.com/NYPL-Simplified/server_core/pull/465 (renaming `Collection` to `Catalog` so I can add a different `Collection` class) doesn't break any metadata wrangler code. I basically do this by renaming 'collection' to 'catalog' everywhere.

Note that the core branch drops the `collections` and `collectionidentifiers` tables so they can be recreated under new names. Since there are no collections currently in the metadata wrangler database, I figure it's safe.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -10,7 +10,7 @@ from core.app_server import (
     URNLookupController as CoreURNLookupController,
 )
 from core.model import (
-    Collection,
+    Catalog,
     CoverageRecord,
     DataSource,
     Identifier,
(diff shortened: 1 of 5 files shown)


controller.py
```diff
@@ -148,19 +148,19 @@ def remove_items(self):
...
-                    collection.catalog.remove(identifier)
+                if identifier in catalog.catalog:
+                    catalog.catalog.remove(identifier)
                     message = OPDSMessage(
                         urn, HTTP_OK, "Successfully removed"
                     )
                 else:
                     message = OPDSMessage(
-                        urn, HTTP_NOT_FOUND, "Not in collection catalog"
+                        urn, HTTP_NOT_FOUND, "Not in catalog catalog"
```
courte: Now only one "category" is needed :)

PR #110, opened 2017-03-22
Catalog against collections

courte:
Uses ClientServer and Collection classes and their underlying assumptions from NYPL-Simplified/server_core#481 to catalog different collections. It adds some endpoints to register and update an authenticated `ClientServer`.

Supports eventual NYPL-Simplified/circulation#467.

### DIFF
--- a/app.py
+++ b/app.py
@@ -108,11 +108,6 @@ def remove(collection_metadata_identifier):
         collection_details=collection_metadata_identifier
     )
 
-@app.route('/client/register', methods=['POST'])
-@returns_problem_detail
-def register_client():
-    return CatalogController(Conf.db).register_client()
-
 @app.route('/client/update_url', methods=['POST'])
 @requires_auth
 @returns_problem_detail
--- a/app.py
+++ b/app.py
@@ -17,6 +17,7 @@ from core.model import production_session
 from core.config import Configuration
 
 from controller import (
+    authenticated_server_from_request,
     CatalogController,
     CanonicalizationController,
     URNLookupController
@@ -48,9 +49,7 @@ else:
 def accepts_auth(f):
     @wraps(f)
     def decorated(*args, **kwargs):
-        server = CatalogController(Conf.db).authenticated_server_from_request(
-            required=False
-        )
+        server = authenticated_server_from_request(Conf.db, required=False)
         if isinstance(server, ProblemDetail):
             return server.response
         return f(*args, **kwargs)
@@ -59,7 +58,7 @@ def accepts_auth(f):
 def requires_auth(f):
     @wraps(f)
     def decorated(*args, **kwargs):
-        server = CatalogController(Conf.db).authenticated_server_from_request()
+        server = authenticated_server_from_request(Conf.db)
         if isinstance(server, ProblemDetail):
             return server.response
         return f(*args, **kwargs)
@@ -83,7 +82,7 @@ def heartbeat():
 def canonical_author_name():
     return CanonicalizationController(Conf.db).canonicalize_author_name()
 
-@app.route('/lookup', defaults=dict(collection_metadata_identifier=None))
+@app.route('/lookup')
 @app.route('/<collection_metadata_identifier>/lookup')
 @accepts_auth
 @returns_problem_detail
--- a/app.py
+++ b/app.py
@@ -17,7 +17,7 @@ from core.model import production_session
 from core.config import Configuration
 
 from controller import (
-    authenticated_server_from_request,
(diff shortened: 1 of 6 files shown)


app.py
```diff
@@ -45,23 +46,21 @@ def initialize(cls, _db):
...
+def accepts_auth(f):
     @wraps(f)
     def decorated(*args, **kwargs):
-        catalog = CatalogController(Conf.db).authenticated_catalog_from_request(
-            required=False
-        )
-        if isinstance(catalog, ProblemDetail):
-            return catalog.response
-        return f(catalog=catalog, *args, **kwargs)
+        server = authenticated_server_from_request(Conf.db, required=False)
```
leonardr: As I mentioned in my core review, referring to this as a 'server' is confusing. From our perspective, it's a client of an API. The fact that it also happens to be a server (a circ manager) doesn't matter in this context.

controller.py
```diff
@@ -129,11 +141,29 @@ def update_url(time=last_update_time, page=None):
...
+        except ValueError, e:
+            return INVALID_INPUT.detailed(e.message)
+
+    def remove_items(self, collection_details):
+        """Removes identifiers from a collection's catalog"""
+        server = authenticated_server_from_request(self._db)
+        if isinstance(server, ProblemDetail):
+            return server
 
+        collection = Collection.from_metadata_identifier(self._db, collection_details)[0]
```
leonardr: I use the `collection, ignore = Collection...` syntax in this situation to avoid giving the impression that `from_metadata_identifier` might return a large number of `Collection` objects and we're just picking the first one.

controller.py
```diff
@@ -159,16 +189,32 @@ def remove_items(self):
                     )
             if message:
                 messages.append(message)
+        self._db.commit()
```
leonardr: The database session should already be committed at the end of each request by `shutdown_session` in `app.py`, so this shouldn't be necessary. If it is, then something's wrong with `app.py.`

controller.py
```diff
@@ -159,16 +189,32 @@ def remove_items(self):
...
+        server = authenticated_server_from_request(self._db)
+        if isinstance(server, ProblemDetail):
+            return server
+
+        url = request.args.get('client_url')
+        if not url:
+            return INVALID_INPUT.detailed("No 'client_url' provided")
+
+        server.url = ClientServer.normalize_url(urllib.unquote(url))
+        self._db.commit()
```
leonardr: Similar comment to last `_db.commit()` call.

controller.py
```diff
@@ -159,16 +189,32 @@ def remove_items(self):
...
+
+        return make_response("", HTTP_ACCEPTED)
```
leonardr: Since the change has been already been made, 200 OK is a better response here.

controller.py
```diff
@@ -235,10 +281,13 @@ def process_urn(self, urn, catalog=None, **kwargs):
...
-        # this Identifier part of the given catalog.
-        if catalog:
-            catalog.catalog_identifier(self._db, identifier)
+        # If a Collection was provided by an authenticated server, this
+        # Identifier is part of the Collection's catalog.
+        server = authenticated_server_from_request(self._db, required=False)
+        if server and collection_details:
+            collection = Collection.from_metadata_identifier(self._db, collection_details)[0]
+            collection.catalog_identifier(self._db, identifier)
+            self._db.commit()
```
leonardr: Another `_db.commit` that shouldn't be necessary.

controller.py
```diff
@@ -129,11 +141,29 @@ def update_url(time=last_update_time, page=None):
 
         return feed_response(update_feed)
 
-    def remove_items(self):
-        catalog = self.authenticated_catalog_from_request()
-        if isinstance(catalog, ProblemDetail):
-            return catalog
+    def register_client(self):
```
leonardr: I'm not comfortable registering clients through an open API call. In the long run I think the simplest thing would be a callback verification process. If you try to register http://foo/ with the metadata wrangler, the metadata wrangler should POST the secret to http://foo/metadata_wrangler_callback (or some similar URL), and only save the registration if the callback succeeds. 

I'm not saying you should implement this now--it's not necessary given the small number of libraries we're starting with, and it might have security implications I haven't considered.

Unless you do want to implement something like that now (in which case we can talk about an implementation), I would take this bit out for now. I think it's simpler to add clients to the metadata wrangler by running the script and communicating the key to the library out-of-band.

leonardr:
Taking the changes to core as given, the main thing here is what to do with `register_client`.

leonardr:
:+1:

PR #113, opened 2017-04-06
Add catalog input endpoint

courte:
Changes to Collections on the circulation manager necessitate a way to update a catalog without using `/lookup`. This branch adds a `/collection/add_items` endpoint that accepts any number of URNs, creating identifiers to catalog for a unique Collection.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -144,8 +144,53 @@ class CatalogController(object):
 
         return feed_response(update_feed)
 
+    def add_items(self, collection_details):
+        """Adds identifiers to a Collection's catalog"""
+        client = authenticated_client_from_request(self._db)
+        if isinstance(client, ProblemDetail):
+            return client
+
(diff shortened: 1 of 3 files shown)


controller.py
```diff
@@ -144,8 +144,54 @@ def update_url(time=last_update_time, page=None):
...
+
+            if not identifier:
+                message = OPDSMessage(
+                    urn, INVALID_URN.status_code, INVALID_URN.detail
+                )
+            else:
+                description = "Already in catalog"
+                if identifier not in collection.catalog:
+                    collection.catalog_identifier(self._db, identifier)
+                    description = "Successfully added"
```
leonardr: I don't think it matters very much, but the response code in this case should be 201 ("Created"), not 200 ("OK").

leonardr:
:+1: with one tiny change.

PR #114, opened 2017-04-25
Use new core

leonardr:
This branch does, hopefully, all the work necessary to make the metadata wrangler support the changes made to core for multi-tenancy. Basically this means changing the CoverageProviders into CollectionCoverageProviders. I resisted the urge to go further and make https://github.com/NYPL-Simplified/metadata_wrangler/pull/100 happen as well.


content_cafe.py
```diff
@@ -27,30 +27,32 @@
...
-        self.content_cafe = api or ContentCafeAPI(
-            self._db, self.mirror, uploader=uploader
-        )
 
-        super(ContentCafeCoverageProvider, self).__init__(
-            "Content Cafe Coverage Provider",
-            self.input_identifier_types, output_source,
-            batch_size=25)
+class ContentCafeCoverageProvider(IdentifierCoverageProvider):
+    SERVICE_NAME = "Content Cafe Coverage Provider"
```
leonardr: Since Content Cafe only provides metadata for individual titles and doesn't have any relationship to the collection that title is in, I made it an IdentifierCoverageProvider instead of a CatalogCoverageProvider

content_server.py
```diff
@@ -1,83 +1,83 @@
...
+        super(LookupClientCoverageProvider, self).__init__(
+            collection, **kwargs
         )
 
+        # Assume that this collection's OPDS server also implements
+        # the lookup protocol.
+        feed_url = collection.external_account_id
+        root = urlparse.urljoin(feed_url, '/')
+        self.lookup_client = self._lookup_client(root)
+        self.importer = self._importer()
```
leonardr: This lets us take the 'content server' integration out of the JSON config. We now perform an OPDS lookup on the server that identifies the Collection (which, in general, will be the content server).

coverage.py
```diff
@@ -43,50 +45,50 @@
...
-        )
-        # Other components don't have INTERNAL_PROCESSING as offering
-        # licenses, but we do, because we're responsible for managing
-        # LicensePools.
-        output_source.offers_licenses=True
-        input_identifier_types = [Identifier.OVERDRIVE_ID, Identifier.ISBN]
+    def __init__(self, collection, uploader=None,
+                 viaf_client=None, linked_data_coverage_provider=None,
+                 overdrive_api_class=OverdriveAPI,
+                 **kwargs):
```
leonardr: We're now taking a lot of arguments that let us plug in mock APIs, which wasn't possible before.

oclc.py
```diff
@@ -750,28 +750,34 @@ def run(self):
                     output.write("\n")
 
 
-class LinkedDataCoverageProvider(CoverageProvider):
+class LinkedDataCoverageProvider(IdentifierCoverageProvider):
```
leonardr: Similar story here as with Content Cafe. We don't care which collection or catalog these identifiers belong to, and it doesn't help us get the data (e.g. by providing credentials) since the service is public.

viaf.py
```diff
@@ -868,11 +868,15 @@ def get_data(self, filename):
...
 
-
-
+    def lookup_by_name(self, sort_name, display_name=None, do_get=None,
+                       known_titles=None):
+        def do_get(*args, **kwargs):
+            return (200, {}, self.get_data("mindy_kaling.xml"))
+        return super(MockVIAFClient, self).lookup_by_name(
+            sort_name, display_name, do_get, known_titles
+        )
```
leonardr: This code was copied from test_viaf.py, where it's used in the context of a specific test. There should be a more useful mock VIAF client that lets us do simpler VIAF tests, but I don't have the attention span to do that work.

PR #115, opened 2017-06-20
Update multi tenant core

courte:
This brings the MetadataWrangler up to date with recent changes to server_core's `multi-tenant` branch. The main changes are from using parameter-heavy `CoverageProvider` to using the `IdentifierCoverageProvider`.

It also creates two default collections -- one that's wrapped around DataSource.INTERNAL_PROCESSING and another for Overdrive. These Collections allow Works to keep being generated during the Identifier resolution process.

Lastly, it adds URI and GUTENBERG identifier types back to the resolution list.


courte:
Turns out this lived in `use-new-core`.

PR #116, opened 2017-06-21
Use external integrations

courte:
This branch imports third-party integrations into the database as ExternalIntegrations and makes sure those objects are used by the relevant APIs.

It also returns the `IdentifierResolutionCoverageProvider` to covering open access identifiers. (I think I removed that a while ago, and I shouldn't have.)


bin/repair/bibliographic_coverage
```diff
@@ -1,12 +0,0 @@
-#!/usr/bin/env python
```
courte: I came into this file to update the use of `S3Uploader`, but then I realized that the `BibliographicRefreshScript` is broken on core. On the Metadata Wrangler, we'd probably prefer to force re-run the IdentifierResolutionCoverageProvider with those particular identifiers, so I deleted it here.

content_cafe.py
```diff
@@ -73,22 +74,35 @@ class ContentCafeAPI(object):
...
+
+    @classmethod
+    def from_config(cls, _db, mirror, **kwargs):
+        integration = ExternalIntegration.lookup(
+            _db, ExternalIntegration.CONTENT_CAFE,
+            ExternalIntegration.METADATA_GOAL
+        )
+
+        if not integration or not (integration.username and integration.password):
+            raise ValueError('Content Cafe not properly configured')
```
leonardr: Can you use CannotLoadConfiguration here? While starting up the circ manager from scratch I decided it's be easier if we changed CannotLoadConfiguration to start referring to configuration items stored in the database, not just in the config file.

coverage.py
```diff
@@ -89,7 +92,8 @@ def __init__(self, collection, uploader=None,
         # Since we are the metadata wrangler, any resources we find,
         # we mirror to S3.
         if not uploader:
-            uploader = S3Uploader()
+            uploader = S3Uploader.from_config(self._db)
```
leonardr: I have a branch that changes `S3Uploader.from_config()` to return None instead of raising an exception if no S3 integration is defined. In the circ manager this is the right thing to do, but in metadata and content you need to check that an S3Uploader is actually returned, and raise `CannotLoadConfiguration` if not.

mirror.py
```diff
@@ -29,7 +29,7 @@ class CoverImageMirror(object):
     def __init__(self, db, uploader=None):
         self._db = db
         self.data_source = DataSource.lookup(self._db, self.DATA_SOURCE)
-        self.uploader = uploader or S3Uploader()
+        self.uploader = uploader or S3Uploader.from_config(self._db)
```
leonardr: Actually, given how many times you're calling this, I think it'd be better if I added a `required` argument to `from_config`, which you could use instead of checking the result yourself.

shadow_catalog.py
```diff
@@ -86,9 +87,19 @@ class ShadowCatalogAPI(object):
...
     # 995: ???
 
+    @classmethod
+    def from_config(cls, _db):
+        integration = ExternalIntegration.lookup(
+            _db, ExternalIntegration.NYPL_SHADOWCAT,
+            ExternalIntegration.METADATA_GOAL
+        )
+        if not integration.url:
+            raise ValueError('No url found for NYPL Shadowcat')
```
leonardr: CannotLoadConfiguration here.

leonardr:
I'm going to make a change to how `S3Uploader.from_config` works in https://github.com/NYPL-Simplified/server_core/pull/539. Once I do that, incorporate that core branch into this and you'll be good to go. Also you can merge use_new_core into multi-tenant, I'm not sure why we didn't do that before.

coverage.py
courte: If the circ manager isn't using an S3 integration, it shouldn't ever be calling this method. How does an error raised in `from_config` get in the way of anything in that app?

leonardr:
I ended up preserving a variant of the old behavior -- if you call `S3Uploader.from_config` with no special arguments and there's no S3 configured, you'll get CannotLoadConfiguration. So no need to change anything now, I think.

coverage.py
leonardr: `Configuration.load()` calls `S3Uploader.integration()`, I think so we can store the bucket names at startup time and not have to keep checking them. 

But, it's true that we don't really need to store those bucket names at startup time. Unlike with CDN URLs we aren't checking the bucket names millions of times in a time-critical part of the code. 

So I think we could take out the reference to S3Uploader in Configuration.load without causing problems. Does that make sense?

courte: Ah, okay, I see. Yeah, your change is right.

PR #120, opened 2017-08-30
Mirror and script fixes

aslagle:
This fixes a couple problems I found while working on mirroring covers from the circulation manager:

- The script to create IntegrationClients was never committing.
- The code to mirror a link wasn't fetching the content of a link if there was already a Representation, and Identifier.add_link creates a Representation without fetching the content if there's a media type.


aslagle:
Okay now this only fixes the generate_client script. I also fixed the value it was printing out for client key.

PR #123, opened 2017-09-05
Added endpoint for receiving metadata from an IntegrationClient.

aslagle:
This is the metadata side of https://github.com/NYPL-Simplified/circulation/pull/660

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -28,6 +28,12 @@ from core.model import (
     get_one,
     get_one_or_create,
 )
+from core.metadata_layer import (
+    Metadata,
+    ContributorData,
+    IdentifierData,
+    LinkData,
+)
 from core.opds import (
     AcquisitionFeed,
     LookupAcquisitionFeed,
@@ -257,13 +263,6 @@ class CatalogController(object):
 
                 message = OPDSMessage(urn, status, description)
 
-                links = entry.get("links")
-                images = [l for l in links if l.get("rel") == Hyperlink.IMAGE or l.get("rel") == Hyperlink.THUMBNAIL_IMAGE]
-
-                for image in images:
-                    link, is_new  = identifier.add_link(image.get("rel"), image.get("href"),
-                                                        data_source=data_source)
-
                 # Make sure there's a LicensePool for this Identifier in this
                 # Collection.
                 license_pools = [p for p in identifier.licensed_through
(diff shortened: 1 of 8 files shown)


controller.py
```diff
@@ -208,6 +214,106 @@ def add_items(self, collection_details):
...
+
+                if identifier not in collection.catalog:
+                    collection.catalog_identifier(self._db, identifier)
+                    status = HTTP_CREATED
+                    description = "Successfully added"
+
+                message = OPDSMessage(urn, status, description)
+
+                links = entry.get("links")
+                images = [l for l in links if l.get("rel") == Hyperlink.IMAGE or l.get("rel") == Hyperlink.THUMBNAIL_IMAGE]
```
leonardr: A lot of this code is reminiscent of code from `metadata_layer.py`. Can you use the OPDSImporter class to turn the OPDS entry into a Metadata object? Or build a Metadata object some other way?

integration_client.py
```diff
@@ -0,0 +1,31 @@
...
+        self.scaler = ImageScaler(_db, [self.mirror], uploader=uploader)
+
+        super(IntegrationClientCoverageProvider, self).__init__(
+            collection, *args, **kwargs)
+
+    def process_item(self, identifier):
+        edition = self.edition(identifier)
+        self.mirror.mirror_edition(edition)
+        self.scaler.scale_edition(edition)
+        return identifier
```
leonardr: I get why you want to do the mirroring in a separate step, instead of right as the OPDS feed is uploaded, but I was hoping to get rid of `ImageScaler` and `CoverImageMirror` in the near future. Could you rewrite this using `Metadata`?

But, maybe this is a good use for `ImageScaler` and `CoverImageMirror` and we should keep them around.

aslagle:
I changed the controller to use `Metadata`. The only problem is the initial request is slower since it does a presentation calculation now, instead of waiting until the coverage provider runs.

aslagle:
And I can probably add a PresentationCalculationPolicy for that.

leonardr:
That looks good and a PresentationCalculationPolicy should defer the performance hit. How about the image mirror thing?

aslagle:
I'm working on that. Metadata.from_edition doesn't get the links, so I think I'll have to change that.

aslagle:
@leonardr this is ready for another look.

PR #124, opened 2017-09-05
Use opds authentication

courte:
This is the first half of allowing OPDS authentication between the circulation manager and the metadata wrangler. (The second half is going to be on the circulation manager, of course.)

Much of the code was pulled heartlessly from the lonely, clutching arms of the Library Registry. It might be nice to put it in a mix-in or module of some sort if we see a lot of apps doing this kind of exchange in the future.


controller.py
```diff
@@ -73,6 +76,66 @@ def authenticated_client_from_request(_db, required=True):
...
+        self._db = _db
+
+    def opds_catalog(self):
+        url = ConfigurationSetting.sitewide(self._db, Configuration.BASE_URL_KEY).value
+        catalog = dict(
+            id=url,
+            title='Library Simplified Metadata Wrangler',
+        )
+
+        catalog['links'] = [
```
courte: This hardcoding... 🙈, but dynamically loading the links would have been about the same, and I wasn't that the data structure to hold this information wouldn't look pretty much exactly like this. I'm open to alternatives.

controller.py
```diff
@@ -238,7 +303,7 @@ def add_with_metadata(self, collection_details):
         entries = feed.get("entries", [])
 
         if not client.data_source:
-            client.data_source = DataSource.lookup(self._db, client.key, autocreate=True)
+            client.data_source = DataSource.lookup(self._db, client.url, autocreate=True)
```
courte: This will mean that the DataSource is tied to the overarching server from which the library edit came -- not to any particular library on that server.

problem_details.py
```diff
@@ -0,0 +1,31 @@
+from core.util.problem_detail import ProblemDetail as pd
```
courte: Stolen wholesale from the library registry.

tests/test_overdrive.py
```diff
@@ -47,13 +47,13 @@ def test_replacement_policy_uses_provided_mirror(self):
         test_cover = self.data_file("covers/test-book-cover.png")
         test_small_cover = self.data_file("covers/tiny-image-cover.png")
 
-        # Overdrive's thumbnail image -- we will not be using this
```
courte: A swap for the new DummyHTTPClient

controller.py
leonardr: One advantage of loading the links with `url_for` is you can make them absolute URLs, which will be easier for a client to handle. Another is the test that generates this file will break if we change anything about the controllers. If it's possible to use `url_for` to make URI Templates then I think you should go for it. But it might not work well.

controller.py
leonardr: How does this situation differ from https://github.com/NYPL-Simplified/metadata_wrangler/pull/123/files#diff-7a315c65deca2c033ce485bfa9de4f4aR225, where we use client.key?

leonardr: Upon further reflection I think I understand why we use client.url here, but now I think we should also use client.url in the other place as well. So either we need another change or I'm confused.

tests/test_controller.py
```diff
@@ -387,29 +427,196 @@ def test_remove_items(self):
...
+        del no_key_json['public_key']['value']
+        no_public_key_response.content = json.dumps(no_key_json)
+        self.http.responses.append(no_public_key_response)
+
+        request_args = self.create_register_request_args(document_url)
+        with self.app.test_request_context('/', **request_args):
+            response = self.controller.register(do_get=self.http.do_get)
+        assert_invalid_auth_document(response, "missing an RSA public_key")
+
+    def test_register_succeeds_with_valid_authentication_document(self):
```
leonardr: If I'm reading this test right, you're using the `public_key` of an authentication document (associated with a particular library) to sign the shared secret for the entire circulation manager. At first I thought this was totally wrong, but upon reflection I see how this works. It doesn't matter which key you use to sign the shared secret with, so long as it's one the circulation manager has access to, and this lets you get rid of the [metadata key document](https://github.com/NYPL-Simplified/Simplified/wiki/Metadata-Wrangler#metadata-key-document) entirely. However, it is a little confusing, because you're mixing library-level stuff with server-level stuff.

controller.py
courte: It's not different from the previous change -- it's just that client.key doesn't exist anymore in this branch. Either way, changes would be server-specific, not library-specific.

controller.py
```diff
@@ -73,6 +76,66 @@ def authenticated_client_from_request(_db, required=True):
...
+                "href": "/lookup{?urn*}",
+                "type": "application/atom+xml;profile=opds-catalog",
+                "title": "Look up metadata about one or more specific items",
+                "templated": "true"
+            },
+            {
+                "rel": "http://opds-spec.org/sort/new",
+                "href": "/{collection_metadata_identifier}/updates",
+                "type": "application/atom+xml;profile=opds-catalog",
+                "title": "Recent changes to your tracked collection",
```
leonardr: I'd change all these `title`s to "one of your tracked collections" to make it clear that you can track multiple collections.

controller.py
```diff
@@ -107,6 +170,8 @@ def canonicalize_author_name(self):
 class CatalogController(object):
     """A controller to manage a Collection's catalog"""
 
+    OPDS_CATALOG_REGISTRATION_MEDIA_TYPE = "application/opds+json;profile=https://librarysimplified.org/rel/profile/directory"
```
leonardr: Since we aren't serving links to OPDS servers, just an ID and the shared secret, you should be able to use the normal OPDS 2 media type.

controller.py
```diff
@@ -380,19 +445,100 @@ def remove_items(self, collection_details):
 
         return feed_response(removal_feed)
 
-    def update_client_url(self):
-        """Updates the URL of a IntegrationClient"""
-        client = authenticated_client_from_request(self._db)
-        if isinstance(client, ProblemDetail):
-            return client
+    def register(self, do_get=HTTP.get_with_timeout):
+        opds_url = request.form.get('url')
```
leonardr: I recently changed the circulation/registry interface so that the URL submitted is the URL to the auth document itself (which must be publically available, no 401 allowed). Whether you keep the auth document in this role or replace it with a new type of document, I recommend making the corresponding change here. It eliminates a lot of special cases.

problem_details.py
leonardr: I've changed these in a recent registry branch. You should be able to eliminate some of them by going direct to the document with the public key in it.

tests/files/controller/auth_document.json
```diff
@@ -0,0 +1,60 @@
+{
```
leonardr: As I suggested on Slack, as a proof of concept you should be able to strip everything out of here except the ID and the public key.

tests/files/controller/circ_feed.opds
```diff
@@ -0,0 +1,163 @@
+<feed xmlns:app="http://www.w3.org/2007/app" xmlns:bibframe="http://bibframe.org/vocab/" xmlns:dcterms="http://purl.org/dc/terms/" xmlns:drm="http://librarysimplified.org/terms/drm" xmlns:opds="http://opds-spec.org/2010/catalog" xmlns:schema="http://schema.org/" xmlns:simplified="http://librarysimplified.org/terms/" xmlns="http://www.w3.org/2005/Atom">
```
leonardr: If you change to pull the document with the public key directly, you won't need this as an intermediary.

controller.py
```diff
@@ -446,64 +445,47 @@ def remove_items(self, collection_details):
...
+            return INVALID_INTEGRATION_DOCUMENT
 
-        url = auth_response.get('id')
+        public_key_response = response.json()
+
+        url = public_key_response.get('id')
         if not url:
-            return INVALID_AUTH_DOCUMENT.detailed(
+            return INVALID_INTEGRATION_DOCUMENT.detailed(
                 "The OPDS authentication document is missing an id."
```
leonardr: You're not expecting to get an authentication document anymore. You should pick a name for the document you are expecting and use that in these messages.

tests/files/controller/auth_document.json
leonardr: This file should be renamed.

PR #125, opened 2017-09-07
Track unauthenticated lookups in a new "unaffiliated" collection

leonardr:
This branch makes sure that all identifiers that come in to be resolved are associated with some collection. Identifiers that come in through unauthenticated lookup are added to a new collection called 'Unaffiliated Identifiers'. 

This guarantees that the `IdentifierResolutionCoverageProvider` will eventually process these identifiers, though it only gets to that collection after it's caught up on all other collections.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -485,7 +485,7 @@ class URNLookupController(CoreURNLookupController):
             return self.add_work(identifier, work)
 
         # Work remains to be done.
-        return self.register_identifier_as_unresolved(urn, identifier, collection)
+        return self.register_identifier_as_unresolved(urn, identifier)
 
     def register_identifier_as_unresolved(self, urn, identifier):
         # This identifier could have a presentation-ready Work
--- a/controller.py
+++ b/controller.py
@@ -49,6 +49,7 @@ from core.problem_details import (
     INVALID_URN,
 )
 
+from coverage import IdentifierResolutionCoverageProvider
 from canonicalize import AuthorNameCanonicalizer
 
 HTTP_OK = 200
@@ -181,7 +182,6 @@ class CatalogController(object):
         collection, ignore = Collection.from_metadata_identifier(
             self._db, collection_details
         )
-
         urns = request.args.getlist('urn')
         messages = []
         for urn in urns:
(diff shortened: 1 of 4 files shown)


leonardr:
Once I deploy this I'll write a script that adds all Identifiers not in a Collection to the unaffiliated collection. Since we only have two instances and not necessary to run the site it's easier to do this afterwards rather than during migration.

controller.py
```diff
@@ -493,6 +492,15 @@ def register_identifier_as_unresolved(self, urn, identifier):
         # work gets done eventually by creating a CoverageRecord
         # representing the work that needs to be done.
         source = DataSource.lookup(self._db, DataSource.INTERNAL_PROCESSING)
+
+        # Add this Identifier to the 'unaffiliated' collection to make
+        # sure it gets covered eventually by the identifier resolution
+        # script, which only covers Identifiers that are in some
+        # collection.
+        collection, ignore = IdentifierResolutionCoverageProvider.unaffiliated_collection(
```
courte: If you add identifiers to the collection here, every unresolved identifier will be marked as unaffiliated, regardless of their attachment to other collections. There's a segment in `process_urn` that catalogs Identifiers in an indicated collection. What do you think of cataloguing unaffiliated Identifiers at that point only if no collection has been identified?

leonardr:
Here's the SQL I'm using. It assumes the unaffiliated collection has already been created.

```
insert into collections_identifiers (collection_id, identifier_id) select c.id, i.id from identifiers i join coveragerecords cr on i.id=cr.identifier_id and status='transient failure' and operation='resolve-identifier' left join collections_identifiers ci on ci.identifier_id=i.id and ci.collection_id is null join collections c on c.name='Unaffiliated Identifiers';
```

controller.py
leonardr: A good idea.

controller.py
```diff
@@ -464,6 +463,7 @@ def process_urn(self, urn, collection_details=None, **kwargs):
         # If a Collection was provided by an authenticated IntegrationClient,
         # this Identifier is part of the Collection's catalog.
         client = authenticated_client_from_request(self._db, required=False)
+        collection = None
```
courte: Doesn't matter either way, but I don't think you need this anymore.

PR #127, opened 2017-09-22
Run providers independently

courte:
"_Let CoverageProviders do their job_," as I always say. This branch does just that and especially moves the slow-going work of creating an ISBN's Work object to the OCLCLinkedDataCoverageProvider.

There's a new IdentifierResolutionRegistrar object to make sure that incoming identifiers get set up for all their coverage favorites, as well as giving them a LicensePool and Collection to get started with. There's a bit o' ugliness in the OverdriveBibliographicCoverageProvider to keep it going with catalog-based coverage instead. There's scripts galore. Perhaps a better question is: what _isn't_ here?


bin/content_server_monitor
```diff
@@ -1,25 +0,0 @@
-#!/usr/bin/env python
```
courte: I can't say for sure, but I don't think this is being used anymore.

controller.py
```diff
@@ -527,12 +548,13 @@ class URNLookupController(CoreURNLookupController):
     IDENTIFIER_REGISTERED = "You're the first one to ask about this identifier. I'll try to find out about it."
     WORKING_TO_RESOLVE_IDENTIFIER = "I'm working to locate a source for this identifier."
     SUCCESS_DID_NOT_RESULT_IN_PRESENTATION_READY_WORK = "Something's wrong. I have a record of covering this identifier but there's no presentation-ready work to show you."
-    
-    OPERATION = CoverageRecord.RESOLVE_IDENTIFIER_OPERATION
```
courte: These were moved to `IdentifierResolutionRegistrar`.

coverage.py
```diff
@@ -240,20 +221,6 @@ def providers(self):
             )
 
         return required, optional
-
-    def items_that_need_coverage(self, identifiers=None, **kwargs):
```
courte: With `preregistered_only`, this alteration is no longer necessary.

coverage.py
```diff
@@ -209,14 +194,11 @@ def providers(self):
             optional = []
             required = [content_cafe, oclc_classify]
             
-        # All books derived from OPDS import against the open-access
+        # All books derived from OPDS import against an open-access
         # content server must be looked up in that server.
-        #
-        # TODO: This could stand some generalization. Any OPDS server
```
courte: With the data_source via request, the lookup is now generalized.

coverage.py
```diff
@@ -361,39 +327,6 @@ def finalize(self, identifier):
             self.generate_edition(identifier)
         self.process_work(identifier)
 
-    def generate_edition(self, identifier):
```
courte: Moved to `LinkedDataCoverageProvider` in oclc.py without alteration.

coverage.py
```diff
@@ -421,27 +354,6 @@ def process_work(self, identifier):
             error_msg = "500; " + "Work could not be calculated for %r" % identifier
             raise RuntimeError(error_msg)
 
-    def resolve_equivalent_oclc_identifiers(self, identifier):
```
courte: No longer required. It turns out `LinkedDataCoverageProvider` was already doing all of this!

bin/content_server_monitor
leonardr: Isn't this is how the metadata wrangler gets the initial information about Gutenberg books, so it can go off to OCLC and do a title-author search to get more information?

I think we can definitely get rid of this script in the medium term, but I don't know how else we'd learn the title/author of `http://www.gutenberg.org/etext/863` right now.

leonardr: Or is this now `LookupClientCoverageProvider`'s job?

controller.py
```diff
@@ -78,6 +81,24 @@ def authenticated_client_from_request(_db, required=True):
...
 
 
+def collection_from_details(_db, client, collection_details):
+    if not (client and isinstance(client, IntegrationClient)):
+        return None
+
+    if collection_details:
+        # A DataSource may be sent for collections with the
+        # ExternalIntegration.OPDS_IMPORT protocol.
+        data_source_name = request.args.get('data_source')
```
leonardr: To reflect this change I think you need to add `{?data_source}` to the appropriate links in the OPDS 2 document generated by `opds_catalog`.

tests/test_controller.py
```diff
@@ -779,9 +811,10 @@ def test_process_urn_isbn(self):
         self.assert_one_message(
             isbn.urn, HTTP_CREATED, self.controller.IDENTIFIER_REGISTERED
         )
-        [record] = isbn.coverage_records
-        eq_(record.exception, self.controller.NO_WORK_DONE_EXCEPTION)
-        eq_(record.status, CoverageRecord.TRANSIENT_FAILURE)
+        assert isbn.coverage_records
+        for cr in isbn.coverage_records:
+            eq_(cr.exception, self.controller.registrar.NO_WORK_DONE_EXCEPTION)
+            eq_(cr.status, CoverageRecord.TRANSIENT_FAILURE)
```
leonardr: Is it worth testing which specific tasks were created as CoverageRecords?

courte: I was hoping I could leave that in the tests for TestIdentifierResolutionRegistrar and just make sure it's doing the bare minimum job here.

bin/content_server_monitor
courte: `DetailedOPDSImporter` doesn't exist in core anymore, so I don't think this script is being used. It would be `LookupClientCoverageProvider`'s job now, but this does beg the question of order. I'm going to make sure that OCLC Classify gives a transient failure status to books that don't have a title and author.

courte: It does. So content server works won't be run through OCLC Classify until we get the bibliographic information from the LookupClientCoverageProvider.

PR #131, opened 2017-10-03
Fix scripts

courte:
Some small tweaks to fix bugs and other script-running annoyances after running them all on alpha.

### DIFF
--- a/content_cafe.py
+++ b/content_cafe.py
@@ -55,6 +55,7 @@ class ContentCafeCoverageProvider(IdentifierCoverageProvider):
             self.content_cafe.mirror_resources(identifier)
             return identifier
         except Exception as e:
+            self.log.error('Coverage error for %r', identifier, exc_info=e)
             return self.failure(identifier, repr(e), transient=True)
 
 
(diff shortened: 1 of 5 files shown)


overdrive.py
```diff
@@ -16,12 +17,12 @@ class OverdriveBibliographicCoverageProvider(
 
     EXCLUDE_SEARCH_INDEX = True
 
-    def __init__(self, uploader, collection, *args, **kwargs):
+    def __init__(self, collection, uploader=None, **kwargs):
```
courte: The provider couldn't run with the `RunCollectionCoverageProvider` without this parameter change.

coverage.py
```diff
@@ -318,13 +319,6 @@ def transform_exception_into_failure(self, error, identifier):
 
     def finalize(self, identifier):
         """Sets equivalent identifiers from OCLC and processes the work."""
-
-        if identifier.type==Identifier.ISBN:
-            # In order to create Works for ISBNs, we first have to
-            # create an edition associated with the ISBN as a primary
-            # identifier. At the moment, this is achieved via OCLC
-            # Linked Data.
-            self.generate_edition(identifier)
```
courte: This work was moved to the `LinkedDataCoverageProvider` in the last PR.

content_cafe.py
```diff
@@ -55,6 +55,7 @@ def process_item(self, identifier):
             self.content_cafe.mirror_resources(identifier)
             return identifier
         except Exception as e:
+            self.log.error('Coverage error for %r', identifier, exc_info=e)
```
courte: There's a recurring AttributeError on alpha, and I'm hoping logging the stacktrace will provide more information.

PR #136, opened 2017-11-01
Batch lookup and register URNs

courte:
This branch pushes off registering an identifier to a script and otherwise tries to work in bulk as much as possible.
  - The IdentifierResolutionRegistrar is now a CatalogCoverageProvider, flitting from collection to collection to prepare identifiers.
  - `URNLookupController.work_lookup` has temporarily been hijacked from the base class in core, so that empty xml feeds can be returned to circulation managers on lookup. The goal is to disincentivize folks from continuing to use previous versions of the circulation manager.
  - Lookup now uses the bulk URN parsing from core (NYPL-Simplified/server_core#699)
  - Coverage record are loaded into the database session in one go (instead of 25 separate visits)

I've walked through the scripts between the wrangler and circ locally with an open access collection, and it worked without incident.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -611,7 +611,7 @@ class URNLookupController(CoreURNLookupController):
         TODO: Return to using the BaseURNLookupController.work_lookup
         once the Metadata Wranger load has improved.
         """
-        urns = urns or flask.request.args.getlist('urn')
+        urns = urns or request.args.getlist('urn')
         response = self.process_urns(urns, **process_urn_kwargs)
         self.post_lookup_hook()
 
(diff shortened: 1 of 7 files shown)


controller.py
```diff
@@ -565,9 +555,16 @@ class URNLookupController(CoreURNLookupController):
     log = logging.getLogger("URN lookup controller")
 
     def __init__(self, _db):
-        self.registrar = IdentifierResolutionRegistrar(_db)
+        self.default_collection_id = None
         super(URNLookupController, self).__init__(_db)
 
+    @property
+    def default_collection(self):
```
courte: This (and the TODO comment below) are prep work for when unaffiliated identifiers are welcome again.

PR #138, opened 2017-11-07
Fix updates feed

courte:
This fixes a small error that kept the updates feeds from building properly.

It also adds ISBN entries to the feed as well. When lookup was returning OPDS, this wasn't as important, but now that the updates feed is the only way to get collection info, ISBNs have to be able to return cover links, even if OCLC doesn't result in a work.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -229,8 +229,8 @@ class CatalogController(object):
         entries = []
         for work in works[:]:
             entry = work.verbose_opds_entry or work.simple_opds_entry
-            entry = etree.fromstring(entry)
             if entry:
+                entry = etree.fromstring(entry)
                 entries.append(entry)
                 works.remove(work)
 
--- a/controller.py
+++ b/controller.py
@@ -289,7 +289,11 @@ class CatalogController(ISBNEntryMixin):
                 entries.append(entry)
                 works.remove(work)
 
-        works = [(work.identifier, work) for work in works]
+        works_for_feed = list()
+        for work in works:
+            identifiers = [lp.identifier for lp in work.license_pools]
+            identifiers = [i for i in identifiers if i in collection.catalog]
+            works_for_feed.append((identifiers[0], work))
 
         # Add ISBN entries.
         isbn_ids = [i.id for i in collection.catalog if i.type==Identifier.ISBN]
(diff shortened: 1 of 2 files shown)


controller.py
```diff
@@ -192,9 +193,63 @@ def canonicalize_author_name(self):
         return make_response(author_name, HTTP_OK, {"Content-Type": "text/plain"})
 
 
-class CatalogController(object):
+class ISBNEntryMixin(object):
+
+    def make_opds_entry_from_metadata_lookups(self, identifier):
```
courte: This method is copied almost verbatim from URNLookupController except for two small bits below.

controller.py
```diff
@@ -192,9 +193,63 @@ def canonicalize_author_name(self):
...
+        else:
+            # All metadata lookups have completed. Create that OPDS
+            # entry!
+            entry = identifier.opds_entry()
+
+        if entry is None:
+            # We can't do lookups on an identifier of this type, so
+            # the best thing to do is to treat this identifier as a
+            # 404 error.
+            return OPDSMessage(
```
courte: Instead of directly adding the resulting message or entry to the precomposed entries, it just returns the message/entry itself.

controller.py
```diff
@@ -229,13 +284,31 @@ def update_url(time=last_update_time, page=None):
         entries = []
         for work in works[:]:
             entry = work.verbose_opds_entry or work.simple_opds_entry
-            entry = etree.fromstring(entry)
             if entry:
+                entry = etree.fromstring(entry)
```
courte: This is the bit that was resulting in errors.

controller.py
```diff
@@ -192,9 +193,63 @@ def canonicalize_author_name(self):
...
+        if unaccounted_for:
+            # At least one metadata lookup has not successfully
+            # completed.
+            names = [x.name for x in unaccounted_for]
+            self.log.info(
+                "Cannot build metadata-based OPDS feed for %r: missing coverage records for %s",
+                identifier,
+                ", ".join(names)
+            )
+            return None
```
courte: Instead of directly adding the resulting message or entry to the precomposed entries, it just returns the message/entry itself. Or None, if there's no entry to be had.

PR #139, opened 2017-11-07
Improve logging

leonardr:
This branch remodels the metadata app.py after the circulation manager's app.py. Instead of having a separate `Conf` object, app-specific configuration such as the database connection is kept on the app object itself. The database connection is initialized on the first request, not as soon as the app starts up.

Most importantly, we now call `LogConfiguration.initialize` during initialization, which sets up logging handlers according to the database configuration, potentially including loggly logs. The list of log destinations is among the first things sent to the log, which makes it easy to see if your configuration is correct.

I've tested this locally with loggly configured in my database and verified that log messages are going into a Loggly `HTTPHandler`. I've also verified that those messages show up on Loggly later (search for "Metadata web app."). Individual HTTP requests also show up in Loggly, but it's hard to see them because there's nothing to distinguish them from requests to the circ manager, except possibly the IP address in `clientHost`.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -1,6 +1,8 @@
 import logging
 from nose.tools import set_trace
 
+from core.config import CannotLoadConfiguration
+
 from core.coverage import (
     CoverageFailure, 
     CatalogCoverageProvider, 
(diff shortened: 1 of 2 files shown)


coverage.py
```diff
@@ -133,9 +135,13 @@ def create_overdrive_api(self, overdrive_api_class):
...
-        if is_new:
-            raise ValueError('Default Overdrive collection has not been configured.')
-        return overdrive_api_class(self._db, collection)
+        try:
+            return overdrive_api_class(self._db, collection)
+        except CannotLoadConfiguration, e:
+            self.log.error(
+                'Default Overdrive collection is not properly configured. No Overdrive work will be done.'
+            )
+            return
```
leonardr: This is an unrelated change I made so that I could run coverage providers without configuring an Overdrive integration.

PR #145, opened 2017-11-21
Return work lookup feeds

courte:
This branch re-enables work lookup to align with changes to the circulation manager import process.

### DIFF
--- a/tests/test_controller.py
+++ b/tests/test_controller.py
@@ -420,15 +420,14 @@ class TestCatalogController(ControllerTest):
         eq_(Edition.UNKNOWN_AUTHOR, author.sort_name)
         eq_("eng", edition.language)
 
-        # Finally, the identifier has a transient failure CoverageRecord so it will
-        # be processed by the identifier resolution script.
+        # Create a successful CoverageRecord for the  ResolutionCoverageProvider
         data_source = DataSource.lookup(self._db, DataSource.INTERNAL_PROCESSING)
-        record = CoverageRecord.lookup(identifier, data_source,
-                                        CoverageRecord.RESOLVE_IDENTIFIER_OPERATION)
-        eq_(CoverageRecord.TRANSIENT_FAILURE, record.status)
-        eq_(self.collection, record.collection)
-
-        record.status = CoverageRecord.SUCCESS
+        record, _ignore = CoverageRecord.add_for(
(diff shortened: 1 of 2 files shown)


tests/test_controller.py
```diff
@@ -420,15 +420,14 @@ def test_add_with_metadata(self):
...
-        # Finally, the identifier has a transient failure CoverageRecord so it will
-        # be processed by the identifier resolution script.
+        # Create a successful CoverageRecord for the  ResolutionCoverageProvider
         data_source = DataSource.lookup(self._db, DataSource.INTERNAL_PROCESSING)
-        record = CoverageRecord.lookup(identifier, data_source,
-                                        CoverageRecord.RESOLVE_IDENTIFIER_OPERATION)
-        eq_(CoverageRecord.TRANSIENT_FAILURE, record.status)
-        eq_(self.collection, record.collection)
-
-        record.status = CoverageRecord.SUCCESS
```
leonardr: Removing this code loses the assertion that the controller itself created a coverage record in a transient failure state. The record is still created, but now it's associated with the collection. The `lookup` call you removed won't find it, but adding self.collection to the `lookup` call will make it show up again.

I don't think the CoverageRecord _needs_ to be associated with the collection, but it makes the `IdentifierResolutionCoverageProvider` simpler and it doesn't seem to affect the controller behavior. If CM 1 uploads a cover for a book, and then CM 2 uploads the same cover for the same book, the second CoverageRecord is never created, so the identifier resolution script doesn't have extra work to do. So I think the current code is fine -- just change the `lookup` call back and have it pass in the `collection`.

courte: Oh, you're right! I didn't realize the controller was creating a resolution CoverageRecord.

I'm don't think manually creating the CoverageRecord is required at all anymore, though. Resolution is processed as identifiers are added to a catalog. If an identifier is new, it will be added to the queue for the IdentifierResolutionRegistrar. If it's not, the work of incorporating metadata into an existing Work and recalculating its presentation could be done by the controller. (Right now the PresentationCalculationPolicy isn't doing anything, but it could be.)

leonardr: That sounds like something to tackle in the next pass along with potentially getting rid of the circulation manager's metadata wrangler reaper CoverageProvider.

PR #148, opened 2017-12-01
Fix recurring coverage errors

courte:
This branch uses `IdentifierCoverageProvider.COUNTS_COVERAGE_FOR_EVERY_COLLECTION` to reason about how CoverageRecords should be created, and incorporates a change in core to fix a recurring error getting open access coverage (NYPL-Simplified/server_core#740).

It also pumps up the requests and tinys3 versions in (unfounded) hopes of fixing a recurring dependency error in Content Cafe coverage that I haven't been able to recreate locally.

I'm also pumping up the sqlalchemy version because we've been at 1.0.6 for a bit, and the current sqlalchemy release is 1.1. I ran both metadata and circulation tests on a branch new database using this sqlalchemy version, and I didn't hit any hiccups.

The python-Levenshtein package removes a fuzzywuzzy package warning.

### DIFF
--- a/content_server.py
+++ b/content_server.py
@@ -32,15 +32,17 @@ class LookupClientCoverageProvider(CatalogCoverageProvider):
         self.DATA_SOURCE_NAME = collection.data_source.name
         super(LookupClientCoverageProvider, self).__init__(collection, **kwargs)
 
+        # The feed URL may be a partial list. Get the root URL.
+        feed_url = collection.external_account_id
+        base_url = urlparse.urljoin(feed_url, '/')
+
         # Assume that this collection's OPDS server also implements
         # the lookup protocol.
-        feed_url = collection.external_account_id
-        root = urlparse.urljoin(feed_url, '/')
-        self.lookup_client = self._lookup_client(root)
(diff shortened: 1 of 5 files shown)


content_server.py
```diff
@@ -22,25 +22,30 @@ class LookupClientCoverageProvider(CatalogCoverageProvider):
...
     OPDS_SERVER_RETURNED_WRONG_CONTENT_TYPE = "OPDS Server served unhandleable media type: %s"
    
     def __init__(self, collection, **kwargs):
         self.DATA_SOURCE_NAME = collection.data_source.name
         super(LookupClientCoverageProvider, self).__init__(collection, **kwargs)
 
+        # The feed URL may be a partial list. Get the root URL.
+        feed_url = collection.external_account_id
+        base_url = urlparse.urljoin(feed_url, '/')
+
```
leonardr: When would this happen?

courte: In the case of instant classics collections, where the external_account_id is set to "http://oacontent.librarysimplified.org/works/lists/Instant%20Classics".

leonardr: Got it.

Issue #149, opened 2017-12-04
Used a scoped session for requests


courte:
The circulation manager already uses scoped sessions and can be used as an example.

PR #150, opened 2017-12-04
Use scoped session #149

courte:
This branch starts using a scoped session for metadata wrangler requests, to take advantage of threading.

It also imports from `flask_babel` instead of the deprecated `flask.ext.babel` and uses lazytext for ProblemDetail detail messages. I realize creating lazytext of untranslated (and sometimes variable) text isn't useful, but since the metadata wrangler isn't user-facing, I hope it's okay for now.

### DIFF
--- a/app.py
+++ b/app.py
@@ -6,7 +6,8 @@ import urlparse
 
 from functools import wraps
 from flask import Flask
-from flask.ext.babel import Babel
+from flask_babel import Babel
+from flask_sqlalchemy_session import flask_scoped_session
 
 from core.app_server import (
     HeartbeatController,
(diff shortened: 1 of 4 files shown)


app.py
```diff
@@ -60,12 +59,6 @@ def initialize_database(autoinitialize=True):
     for logger in logging.getLogger().handlers:
         app.log.info("Logs are going to %r", logger)
 
-    # Workaround for a "Resource temporarily unavailable" error when
```
courte: The socket needs to be set before the database is initialized. This has been moved into the `run` method.

controller.py
```diff
@@ -571,7 +572,7 @@ def register(self, do_get=HTTP.get_with_timeout):
         if not url:
             message = "The public key integration document is missing an id."
             log.error(message)
-            return INVALID_INTEGRATION_DOCUMENT.detailed(message)
+            return INVALID_INTEGRATION_DOCUMENT.detailed(_(message))
```
leonardr: The _() should be on the actual text of the message so that it can be pulled out by the translation tool. So `message = _("The public key integration document is missing an id.")`

controller.py
```diff
@@ -609,7 +610,7 @@ def base_url(full_url):
             )
         except ValueError as e:
             log.error("Error in IntegrationClient.register", exc_info=e)
-            return INVALID_CREDENTIALS.detailed(repr(e))
+            return INVALID_CREDENTIALS.detailed(_(repr(e)))
```
leonardr: This one I don't think is even worth localizing -- we don't know what the string is to localize.

leonardr:
:+1:

Issue #151, opened 2017-12-06
Remove duplicate CoverageRecords and IntegrationClient-based datasources.


courte:
There are a lot of duplicate CoverageRecords from the time when the OA content server offered per-collection coverage. There was also a brief period when IntegrationClient's had data_sources that were used to offer coverage / licensepools, and it's all still floating around in the database.

It may be more trouble than its worth right now, but it would be nice to remove the IntegrationClient datasources entirely (and connect their dependents to Collection-based datasources or whatever is in-line with the current state of the code instead). My fear is that inconsistencies will create painful or confusing debugging later on.

PR #152, opened 2017-12-15
Update add with metadata

courte:
This branch updates the `CatalogController.add_with_metadata` endpoint to depend on the IdentifierResolutionRegistrar. If an identifier is added to a new collection's catalog, then the IdentifierResolutionRegistrar will handle the creation of a LicensePool and CoverageRecord; the controller doesn't need to worry about it.

While I was doing that, I started thinking about the flow of an identifier as metadata is added. It's possible that an identifier is "resolved" on the whole, but that metadata is added via some collection *after* that point. So now there's a coverage provider to recalculate the work presentation and a way to incorporate it into IdentifierCoverageProvider subclasses.

Requires NYPL-Simplified/server_core#749.

### DIFF
--- a/bin/work_presentation_coverage
+++ b/bin/work_presentation_coverage
@@ -1,0 +1,12 @@
+#!/usr/bin/env python
+"""Gather information from OCLC Linked Data."""
+import os
+import sys
+bin_dir = os.path.split(__file__)[0]
+package_dir = os.path.join(bin_dir, "..")
+sys.path.append(os.path.abspath(package_dir))
+
+from integration_client import WorkPresentationCoverageProvider
+from core.scripts import RunWorkCoverageProviderScript
+
+RunWorkCoverageProviderScript(WorkPresentationCoverageProvider).run()
--- a/bin/work_presentation_coverage
+++ b/bin/work_presentation_coverage
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-"""Gather information from OCLC Linked Data."""
+"""Calculate Work presentation for registered works."""
 import os
 import sys
 bin_dir = os.path.split(__file__)[0]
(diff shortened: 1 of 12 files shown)


controller.py
```diff
@@ -393,15 +393,15 @@ def add_with_metadata(self, collection_details):
...
+        data_source = DataSource.lookup(
+            self._db, collection.name, autocreate=True
+        )
+
         messages = []
 
         feed = feedparser.parse(request.data)
         entries = feed.get("entries", [])
 
-        if not client.data_source:
```
courte: This is part of the problem detailed in #151.

integration_client.py
```diff
@@ -3,14 +3,126 @@
...
 )
-from core.coverage import CatalogCoverageProvider
 from core.metadata_layer import (
     Metadata,
     ReplacementPolicy,
 )
 
-class IntegrationClientCoverageProvider(CatalogCoverageProvider):
+
+class WorkPresentationCoverageProvider(WorkCoverageProvider):
```
courte: This isn't the ideal file for these classes, but for now I like having the IntegrationClientCoverageProvider separate and this avoids looped imports with `coverage.py`.

coverage.py
```diff
@@ -324,36 +327,9 @@ def transform_exception_into_failure(self, error, identifier):
...
 
-    def finalize(self, identifier):
-        """Sets equivalent identifiers from OCLC and processes the work."""
-        self.process_work(identifier)
-
-    def process_work(self, identifier):
-        """Fill in VIAF data and cover images where possible before setting
-        a previously-unresolved identifier's work as presentation ready.
-
-        TODO: I think this should be split into a separate
```
courte: 🙂

bin/work_presentation_coverage
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python
+"""Gather information from OCLC Linked Data."""
```
leonardr: Docstring needs to be changed.

PR #153, opened 2017-12-20
Add endpoint to request metadata from the circulation manager

### LINKED ISSUE #147: Request metadata info for Axis360 and Bibliotheca books
Create an endpoint that responds with catalog identifiers that need additional metadata from a source on the circulation manager.

courte:
This branch adds an endpoint that finds all of the identifiers in a given collection that have had trouble with resolution and returns them as a paginated feed of OPDS messages. Eventually, the circulation manager will be able to check this endpoint and send helpful data to the wrangler.

I also took the time to refactor some of the code that exists in other parts of the CatalogController around pagination and a bit of the test code.

Fixes #147.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -3,6 +3,10 @@ from datetime import datetime
 from flask import request, make_response
 from flask_babel import lazy_gettext as _
 from lxml import etree
+from sqlalchemy import (
+    and_,
+    not_,
+)
 from sqlalchemy.orm import joinedload
 from Crypto.PublicKey import RSA
 from Crypto.Cipher import PKCS1_OAEP
@@ -62,6 +66,7 @@ from coverage import (
     IdentifierResolutionRegistrar,
 )
 from canonicalize import AuthorNameCanonicalizer
+from integration_client import IntegrationClientCoverageProvider
 from problem_details import *
 
 HTTP_OK = 200
@@ -69,6 +74,7 @@ HTTP_CREATED = 201
 HTTP_ACCEPTED = 202
 HTTP_UNAUTHORIZED = 401
 HTTP_NOT_FOUND = 404
+HTTP_UNPROCESSIBLE_ENTITY = 422
 HTTP_INTERNAL_SERVER_ERROR = 500
 
 OPDS_2_MEDIA_TYPE = 'application/opds+json'
@@ -149,6 +155,12 @@ class IndexController(object):
                 "rel": "http://librarysimplified.org/rel/metadata/collection-remove",
                 "href": "/{collection_metadata_identifier}/remove{?data_source,urn*}",
                 "title": "Remove items from one of your tracked collections.",
+                "templated": "true"
+            },
+            {
+                "rel": "http://librarysimplified.org/rel/metadata/collection-metadata-requests",
+                "href": "/{collection_metadata_identifier}/metadata_requests",
+                "title": "Get items in your collection for which the metadata wrangler needs more information to process.",
                 "templated": "true"
             },
             {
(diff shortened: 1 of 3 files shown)


controller.py
```diff
@@ -62,13 +66,15 @@
...
 from canonicalize import AuthorNameCanonicalizer
+from integration_client import IntegrationClientCoverageProvider
 from problem_details import *
 
 HTTP_OK = 200
 HTTP_CREATED = 201
 HTTP_ACCEPTED = 202
 HTTP_UNAUTHORIZED = 401
 HTTP_NOT_FOUND = 404
+HTTP_UNPROCESSIBLE_ENTITY = 422
```
courte: I don't like using this HTTP code for metadata requests (especially since I think we use it for invalid URNs in the URNLookupController), but I couldn't find another http code that resonated. At the moment, I believe it's going to be ignored on the circulation manager anyway.

controller.py
```diff
@@ -401,74 +431,139 @@ def add_with_metadata(self, collection_details):
...
 
-        for entry in entries:
-            urn = entry.get('id')
-            try:
-                identifier, ignore = Identifier.parse_urn(
-                    self._db, urn
-                )
-            except Exception as e:
-                identifier = None
+        identifiers_by_urn, invalid_urns = Identifier.parse_urns(
```
courte: I wanted to start using bulk parsing here, since we've changed it in other places in the controller and metadata requests may result in more engagement here.

controller.py
```diff
@@ -401,74 +431,139 @@ def add_with_metadata(self, collection_details):
...
-                edition, ignore = metadata.edition(self._db)
-                metadata.apply(edition, collection, replace=replace)
+        for urn in invalid_urns:
+            messages.append(OPDSMessage(
+                urn, INVALID_URN.status_code, INVALID_URN.detail
+            ))
+
+
+        for urn, identifier in identifiers_by_urn.items():
+            entry = entries_by_urn[urn]
```
courte: After this point, the rest of the diff in this method is whitespace changes (de-indentation).

controller.py
leonardr: Yeah, it's tough because this isn't even a real HTTP response. You could send 202 (which implies the server is still on the job) and since the caller knows that they asked for a list of items that it might be able to help with, it will know what 202 means in this context.

controller.py
```diff
@@ -151,6 +157,12 @@ def opds_catalog(self):
                 "title": "Remove items from one of your tracked collections.",
                 "templated": "true"
             },
+            {
+                "rel": "http://librarysimplified.org/rel/metadata/collection-metadata-requests",
+                "href": "/{collection_metadata_identifier}/metadata_requests",
+                "title": "Get items in your collection for which the metadata wrangler needs more information to process.",
+                "templated": "true"
+            },
```
leonardr: I would like to see a term like "missing" or "needs action" in here. With "requests" it's not clear who is making the request, and I find it confusing because the entity making the "request" for metadata is not the entity making the HTTP "request".

controller.py
```diff
@@ -401,74 +431,139 @@ def add_with_metadata(self, collection_details):
...
+        is_awaiting_metadata = self._db.query(Identifier.id)\
+            .join(Identifier.coverage_records).filter(
+                CoverageRecord.data_source_id==data_source.id,
+                CoverageRecord.status==CoverageRecord.REGISTERED,
+                CoverageRecord.operation==IntegrationClientCoverageProvider.OPERATION,
+            ).subquery()
+
+        unresolved_identifiers = unresolved_identifiers.filter(
+            not_(Identifier.id.in_(is_awaiting_metadata))
+        )
```
leonardr: I'm concerned about the performance of this query when either the number of unresolved identifiers or the number of identifiers awaiting metadata is large. I think you could do an outer join against (coverage_records on data_source_id=data_source.id and status=REGISTERED and operation=OPERATION), and restrict the results to fields where that outer join doesn't turn up a record. Something like that should work.

leonardr:
This works well enough that you could merge it now, but there are some naming and performance issues that should be addressed before we start using it for real.

PR #155, opened 2017-12-22
Use consistent timestamp format

leonardr:
This branch fixes a subtle inconsistency in CatalogController.updates_feed, where the incoming timestamp was expected in a certain format, but the outgoing timestamp used in links was a slightly different format. This meant the circulation manager would successfully import the first page of the updates feed, and then the metadata wrangler would crash when the circulation manager asked for the second page.

This branch also gets rid of the crash if a bad timestamp should be sent to the metadata wrangler for whatever reason.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -258,6 +258,8 @@ class CatalogController(ISBNEntryMixin):
     # overall without impacting non-ISBN collections too much.
     UPDATES_SIZE = 35
 
+    TIMESTAMP_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
+
     def __init__(self, _db):
         self._db = _db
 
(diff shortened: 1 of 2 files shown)


controller.py
```diff
@@ -355,10 +365,13 @@ def updates_feed(self, collection_details):
                 entries.append(entry)
 
         title = "%s Collection Updates for %s" % (collection.protocol, client.url)
-
         url_params = dict()
         if last_update_time:
-            url_params = dict(last_update_time=last_update_time)
```
leonardr: This was the bad line. putting the `datetime` object into `url_for` without converting it to a string uses the default string format, which looks like TIMESTAMP_FORMAT but is missing the 'T' and the 'Z'.

tests/test_controller.py
```diff
@@ -339,7 +339,7 @@ def test_updates_feed(self):
                 u"%s Collection Updates for %s" % (self.collection.protocol, self.client.url))
 
             # The timestamp is included in the url.
-            linkified_timestamp = time.strftime("%Y-%m-%d+%H:%M:%S").replace(":", "%3A")
```
leonardr: This test checked a different format from the one on line 329.

leonardr:
Feel free to merge this branch yourself if you like it, since I'll be on vacation next week.

PR #158, opened 2018-01-24
Dramatically improve the performance of remove_items on a large collection.

leonardr:
This branch uses a single SELECT and a single DELETE statement to replace a number of `in` and `remove` statements which were using the database inefficiently. I don't know exactly how bad the situation is, but on a collection the size of NYPL's, that first `in` statement took up enough time all on its own to make a request time out.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -27,6 +27,7 @@ from core.app_server import (
 from core.config import Configuration
 from core.model import (
     Collection,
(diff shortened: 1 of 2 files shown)


leonardr:
Yes, that looks like it has the same problem. I'll do a follow-up branch.

PR #159, opened 2018-01-24
Bring the fast-subset code used in remove_items to add_items

leonardr:
This branch builds on https://github.com/NYPL-Simplified/metadata_wrangler/pull/158/ and also fixes a bug I introduced, where asking to remove an identifier from one catalog would remove it from _all_ catalogs.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -666,8 +666,8 @@ class CatalogController(ISBNEntryMixin):
         # Extract the identifier IDs from the dictionary.
         identifier_ids = [x.id for x in identifiers_by_urn.values()]
 
-        # Find the subset of identifiers that are in the catalog, so
-        # we know to give them a 200 message after deletion.
+        # Find the IDs for the subset of identifiers that are in the
+        # catalog.
         table = collections_identifiers.c
         identifier_match_clause = and_(
             table.identifier_id.in_(identifier_ids),
--- a/controller.py
+++ b/controller.py
@@ -8,6 +8,7 @@ from sqlalchemy import (
     not_,
 )
 from sqlalchemy.orm import joinedload
+from sqlalchemy.sql.expression import and_
 from Crypto.PublicKey import RSA
 from Crypto.Cipher import PKCS1_OAEP
 import base64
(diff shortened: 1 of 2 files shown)


controller.py
```diff
@@ -638,6 +645,37 @@ def remove_items(self, collection_details):
...
+        """
+        # Extract the identifier IDs from the dictionary.
+        identifier_ids = [x.id for x in identifiers_by_urn.values()]
+
+        # Find the subset of identifiers that are in the catalog, so
+        # we know to give them a 200 message after deletion.
+        table = collections_identifiers.c
+        identifier_match_clause = and_(
+            table.identifier_id.in_(identifier_ids),
+            table.collection_id == collection.id
```
leonardr: The absence of this line was what caused the bug in my earlier branch.

courte: 👌

controller.py
```diff
@@ -638,6 +645,37 @@ def remove_items(self, collection_details):
...
+        match_clause). `in_catalog_ids` is a set of Identifier IDs
+        representing the subset of identifiers currently in the
+        catalog. `match_clause` is the SQLAlchemy clause that was used
+        to look up the matching subset.
+        """
+        # Extract the identifier IDs from the dictionary.
+        identifier_ids = [x.id for x in identifiers_by_urn.values()]
+
+        # Find the subset of identifiers that are in the catalog, so
+        # we know to give them a 200 message after deletion.
```
courte: This docstring needs to be updated.

controller.py
```diff
@@ -406,12 +407,24 @@ def add_items(self, collection_details):
...
+        already_in_catalog, ignore = self._in_catalog_subset(
+            collection, identifiers_by_urn
+        )
 
-            if identifier not in collection.catalog:
-                collection.catalog_identifier(identifier)
+        # Everything else needs to be added to the catalog.
+        needs_to_be_added = [
+            x for x in identifiers_by_urn.values()
+            if x not in already_in_catalog
```
courte: If `already_in_catalog` is a list of ids, I think this clause should be `if x.id not in already_in_catalog`... but it appears to work in the tests?

leonardr: The test didn't catch it because the test checks the messages, and when it came time to generate the messages, the code correctly checked the identifier IDs. This method still gave the right results, but some useless work was done because `catalog_identifiers` was called on all the identifiers, even the ones already in the catalog.

PR #163, opened 2018-03-22
Add docker

courte:
Dockerize the metadata wrangler, including support for docker-compose.

Most of these changes are nearly identical to NYPL-Simplified/circulation-docker, with a few key differences that keep things moving:
  - the name of the database and application directory created (though this is largely cosmetic)
  - the addition of node/npm in `simplified_app.sh` (apt-get and api installation)
  - the app module in `uwsgi.ini`
  - the scripts in `docker/services/simplified_crontab`

I hate the amount of duplication -- which ensures that changes beneficial to `circulation-docker` won't make their way here -- but I don't have time to fix it at the moment.

There's also this annoying thing where you can set `depends_on` in `docker-compose`, but it doesn't actually make sure the supporting container enters a "ready" state before it starts the dependent container. There's more info on how to fix it here: https://docs.docker.com/compose/startup-order/.

So with the database and two dependent containers competing to initialize it, you must run `docker-compose up` three times to get both applications to start. The first time, the database is created while the application containers error out. The second time, the database is initialized by one container while the other errors out. The third time -- now fully initialized -- all three containers begin.

### DIFF
--- a/docker/services/simplified_crontab
+++ b/docker/services/simplified_crontab
@@ -1,0 +1,27 @@
+# /etc/cron.d/circulation: Library Simplified Circulation Manager crontab
+# You don't have to run the `crontab' command to install the new
+# version when you edit this file in /etc/cron.d. Files in this directory
+# also have username fields, similar to the systemwide /etc/crontab.
+
+SHELL=/bin/sh
+PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
+HOME=/var/www/circulation
+
+# m h dom mon dow user command
+
+# Register identifiers to for coverage from all third-party sources.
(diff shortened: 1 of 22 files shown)


docker/services/simplified_crontab
```diff
@@ -0,0 +1,27 @@
+# /etc/cron.d/circulation: Library Simplified Circulation Manager crontab
+# You don't have to run the `crontab' command to install the new
+# version when you edit this file in /etc/cron.d. Files in this directory
+# also have username fields, similar to the systemwide /etc/crontab.
+
+SHELL=/bin/sh
+PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
+HOME=/var/www/circulation
```
leonardr: Should be /var/www/metadata

PR #170, opened 2018-05-25
Don't even parse URNs for identifier types we can't resolve.

leonardr:
The metadata wrangler is cluttered with identifiers of types we can't actually resolve. Circulation managers are supposed to only send identifiers that can be resolved, but it happens anyway. This branch makes the metadata wrangler simply refuse to deal with identifiers that can't be resolved. They won't even make it to the database.


controller.py
```diff
@@ -849,27 +855,7 @@ def presentation_ready_work_for(self, identifier):
...
-        ):
-            return True
-        
-        # We can resolve any identifier that's associated with a
-        # presentation-ready work, since the resolution has already
-        # been done--no need to speculate about how.
-        work = self.presentation_ready_work_for(identifier)
-        if work is None:
-            return False
-        return True
```
aslagle: It seems like this would still work, but maybe it’s not necessary.

PR #175, opened 2018-06-08
Merge IdentifierResolutionRegistrar and IdentifierResolutionCoverageProvider

leonardr:
This branch merges two CoverageProviders which did very similar things:

* IdentifierResolutionRegistrar recorded that certain Identifiers needed 'resolution' by calling `register` on a number of other CoverageProviders. It also made sure each such Identifier had a LicensePool. It was invoked from the URNLookupController.
* IdentifierResolutionCoverageProvider went over those same Identifiers and the same CoverageProviders, but it called `ensure_coverage` instead of `register`. That is, it actually did the work instead of telling the individual CoverageProviders to do the work later. It was invoked by the `identifiers_resolve` and `identifiers_resolve_single` script, but technically it was redundant, because the work would be done eventually by the other CoverageProviders.

The new CoverageProvider is called `IdentifierResolutionCoverageProvider`, and depending on how it's configured it will handle an Identifier either by registering all the work that needs to be done (like the old IdentifierResolutionRegistrar did) or by doing all the work immediately (like the old IdentifierResolutionCoverageProvider did). If it notices that any of its CoverageProviders returned a CoverageRecord with the "success" status, it tries to create a presentation-ready Work for the Identifier it was given (CoverageProviders are supposed to do that on their own, but this provides some resiliency).

Basically, the old strategy was to do a bunch of work and create a presentation-ready Work (sometimes) at the end. The new startegy is to create a Work as soon as possible and refine it over time.

This change required that the service-specific CoverageProviders be modified to fit. The problem is that a number of the service-specific CoverageProviders don't work anymore or aren't providing much value given all the processing power they take up. So I got rid of the CoverageProviders that don't work anymore, temporarily disabled the ones that need work, and converted the rest to `CollectionCoverageProviders` that can be handled by IdentifierRegistrationCoverageProvider.

The OverdriveBibliographicCoverageProvider and ContentCafeCoverageProvider are fully integrated into this new system. The IntegrationClientCoverImageCoverageProvider is integrated into this new system, though I'm not 100% sure it will work in a real scenario -- that's next for me to investigate. Here's the fate of the other coverage providers:

* The LookupClientCoverageProvider is removed. This was designed for copying data from the content server to the metadata wrangler. This was always somewhat redundant, and now that the content server is gone, there's nothing for it to do. We can get some of the functionality back (e.g. mirroring free book covers to covers.nypl.org) by improving the IntegrationClientCoverImageCoverageProvider.
* The OCLCClassifyCoverageProvider is disabled -- most likely temporarily, maybe permanently. This which was used to do title/author lookups on books which don't have an ISBN (especially from Project Gutenberg). A title/author lookup is still theoretically useful, but we don't really use Project Gutenberg books anymore and it should be easier to do this with an [https://platform.worldcat.org/api-explorer/apis/wcapi/Bib/SRU](SRU search).
* The LinkedDataCoverageProvider is temporarily disabled. This is a really useful tool for getting classifications and descriptions, but it's processor and IO-intensive, and it turns out not to be all that necessary, since all currently supported license sources either provide decent descriptions and classifications, or (Bibliotheca) contain only books that can be covered by Content Cafe.

Other changes:

* Titles identified by ISBNs weren't reliably getting Works. This meant that even after the metadata wrangler had handled them, their OPDS entries were being cobbled together at runtime from Representation lookups. This both hurt performance and led to us not showing all the information about these titles that we had available. This is a big problem because most data sources identify books using ISBNs.

So now, a presentation-ready Work is created as soon as _any_ coverage provider runs, so long as we know the book's title. (Even this restriction can be relaxed in the future.)  The goal here is to get a description and scaled cover image for a book almost immediately, rather than waiting for the lengthy OCLC Linked Data crawl to finish.

* Coverage providers that add contributor information will, upon completion, make sure the Contributors of the presentation-ready Work have their VIAF information updated. Currently the only provider that does this is the OverdriveBibliographicCoverageProvider, but the LinkedDataCoverageProvider and OCLCClassifyCoverageProvider will do so once they're fixed, and the IntegrationClientCoverImageCoverageProvider might do so in the future. This might pose a problem with the same Contributors having their VIAF information updated multiple times, but I can deal with that when it starts happening.

* Since we no longer treat ISBNs differently, CatalogController.updates_feed could be simplified significantly. Identifiers of all types are expected to get presentation-ready Works as soon as possible.

In addition to the unit tests, I've tested this manually with ISBNs and Overdrive IDs, and tried various things like deleting a successfully created Work and trying the lookup again.


content_cafe.py
```diff
@@ -33,7 +35,9 @@
 
 from core.mirror import MirrorUploader
 
-class ContentCafeCoverageProvider(BibliographicCoverageProvider):
+from coverage_utils import MetadataWranglerBibliographicCoverageProvider
+
+class ContentCafeCoverageProvider(MetadataWranglerBibliographicCoverageProvider):
```
leonardr: This is just a BibliographicCoverageProvider that overrides `work` to create an INTERNAL_PROCESSING dummy LicensePool if necessary rather than creating a LicensePool that comes from the DataSource of the CoverageProvider (which in this case would be Content Cafe).

content_cafe.py
```diff
@@ -45,27 +49,31 @@ class ContentCafeCoverageProvider(BibliographicCoverageProvider):
...
     DATA_SOURCE_NAME = DataSource.CONTENT_CAFE
     
-    def __init__(self, collection, api=None, mirror=None, **kwargs):
+    def __init__(self, collection, api=None, replacement_policy=None, **kwargs):
         """Constructor.
 
         :param collection: A Collection.
         :param api: A ContentCafeAPI.
-        :param mirror: A MirrorUploader.
+        :param replacement_policy: A ReplacementPolicy.
```
leonardr: Every constructor that use to take a Mirror now takes a complete ReplacementPolicy.

content_cafe.py
```diff
@@ -45,27 +49,31 @@ class ContentCafeCoverageProvider(BibliographicCoverageProvider):
...
+                mirror=mirror
+            )
+
         # We pass in registered_only=True because we don't need to cover
         # every single ISBN in the system (most of which are alternate
         # ISBNs found in OCLC Linked Data), only the ISBNs that a
         # client specifically asked to look up.
         super(ContentCafeCoverageProvider, self).__init__(
-            collection=collection, registered_only=True, **kwargs
+            collection=collection, registered_only=True,
```
leonardr: I think I can remove this `registered_only` thing -- alternate ISBNs found in OCLC Linked Data won't be cataloged in any collection.

controller.py
```diff
@@ -211,58 +210,7 @@ def canonicalize_author_name(self):
         return make_response(author_name, HTTP_OK, {"Content-Type": "text/plain"})
 
 
-class ISBNEntryMixin(object):
```
leonardr: This whole thing is gone because ISBNs will reliably get works now. (Though we may need some kind of migration to create Works for ISBNs that have already been processed but didn't get Works.)

controller.py
```diff
@@ -814,12 +754,9 @@ def base_url(full_url):
...
 
 
-class URNLookupController(CoreURNLookupController, ISBNEntryMixin):
+class URNLookupController(CoreURNLookupController):
 
-    UNRESOLVABLE_IDENTIFIER = "I can't gather information about an identifier of this type."
-    IDENTIFIER_REGISTERED = "You're the first one to ask about this identifier. I'll try to find out about it."
-    WORKING_TO_RESOLVE_IDENTIFIER = "I'm working to locate a source for this identifier."
-    SUCCESS_DID_NOT_RESULT_IN_PRESENTATION_READY_WORK = "Something's wrong. I have a record of covering this identifier but there's no presentation-ready work to show you."
+    WORKING_TO_RESOLVE_IDENTIFIER = "I don't have enough information about this Identifier yet.\nDetailed work log:\n "
```
leonardr: This architecture dramatically cuts down on the possible `/lookup` responses -- either there's a presentation-ready work or there isn't. If there isn't, we provide a record of what work is registered for the given Identifier. If there is, there might be more work to be done later, but at least we have some information about the Identifier.

scripts.py
```diff
@@ -519,86 +519,6 @@ def fix_identifier_with_equivalents(self, primary_identifier, equivalent_ids):
         print "I WOULD NOW EXPECT EVERYTHING TO BE FINE."
 
 
-class RedoOCLCForThreeMScript(Script):
```
leonardr: This script is irrelevant now (metadata wrangler doesn't even keep track of Bibliotheca identifiers anymore) and the tests were failing, so I got rid of the whole thing.

controller.py
aslagle: If there is more work to be done later, the client won't know? Or will it know by the absence of some of the information it wanted?

controller.py
```diff
@@ -901,39 +843,66 @@ def process_urns(self, urns, collection_details=None, **kwargs):
...
+
+        :param identifier: The Identifier that needs to be processed.
+        :param urn: The original URN provided by the client. This
+            might be different from Identifier.urn, e.g. because of
+            ISBN normalization.
+        :param collection: The Identifier was registered with this collection.
+        :param resolver: An IdentifierResolutionCoverageProvider which
+            will either create a presentation-ready Work immediately, or
+            make sure that one eventually gets created.
+        :param return: None.
```
aslagle: that's not a param

controller.py
leonardr: The client won't know, but there's nothing it could do differently if it did know.
Overall we need better visibility into where the bottlenecks and failure points are in coverage, but on the level of individual items I think most people won't care once we get the basics -- cover image and description.

coverage_utils.py
```diff
@@ -0,0 +1,68 @@
...
+                identifier, license_pool
+            )
+
+
+class ResolveVIAFOnSuccessCoverageProvider(MetadataWranglerBibliographicCoverageProvider):
+    """A mix-in class for metadata wrangler BibliographicCoverageProviders
+    that add author information. When such a coverage provider
+    completes its work, it should run any Contributors associated with
+    the presentation Edition through VIAF. Then it should try to
+    create a presentation-ready work.
```
aslagle: Why not make VIAF a separate coverage provider? Other coverage providers could tell it to run immediately or not, and then it could keep a separate record of work that was done. I guess it should be a ContributorCoverageProvider rather than an IdentifierCoverageProvider, which we probably don't have yet.

leonardr: This is a good idea.

The whole VIAF thing is another issue that turned out to be kind of marginal. Like OCLC Lookup, we overestimated its importance because we started out dealing with books which had nearly no metadata. The books we deal with now don't have as many problems. So I'm not too worried about how well VIAF lookup works until we have a way of indicating that two Contributor objects refer to the same person.

PR #177, opened 2018-06-11
Create a Work for identifiers we have some information about, even if that information doesn't include the title.

leonardr:
This branch deals with the aftermath of https://jira.nypl.org/browse/SIMPLY-332

A number of scripts are removed -- the underlying functionality was removed earlier. I left in place scripts that we want to have eventually, but right now -- I'll comment them out in the crontab.


coverage.py
```diff
@@ -333,7 +335,9 @@ def process_item(self, identifier):
             # CoverageProvider didn't try to create a Work, or that a
             # preexisting Work has been removed. In the name of
             # resiliency, we might as well try creating a Work.
-            work, is_new = license_pool.calculate_work(even_if_no_author=True)
+            work, is_new = license_pool.calculate_work(
+                even_if_no_author=True, even_if_no_title=True
+            )
```
leonardr: This is the main improvement that actually helped things. Now, looking up an Identifier that's had all its work done, but doesn't have a Work, will create the Work. However, we still need to make sure all those Identifiers get Works ahead of time, because creating a Work can be very expensive in certain cases.

coverage.py
```diff
@@ -370,7 +374,7 @@ def process_one_provider(self, identifier, provider):
                 identifier, force=self.force
             )
         else:
-            coverage_record = provider.register(
+            coverage_record, is_new = provider.register(
```
leonardr: This was an error hidden by the fact that we don't test this with real CoverageProviders.

PR #181, opened 2018-06-26
Rename book-covers.nypl.org URLs to covers.nypl.org

leonardr:
This branch changes the metadata wrangler's dataset to reflect the fact that all old cover images have been moved to the covers.nypl.org S3 bucket, and that there's no longer any need to refer to book-covers.nypl.org.

We could run the same migration script on every circulation manager, but that would mean a big one-time chunk of downtime. Since it's not urgent that we get rid of book-covers.nypl.org, for now I'm okay with changing the metadata wrangler and having the information gradually filter out to the circulation managers as metadata updates.

I'll attach the script I used to generate the complex "replace()" call in the SQL statement.

This should complete the work of https://jira.nypl.org/browse/SIMPLY-285.


leonardr:
```
prefixes = ["http://", "https://"]
replace = [
    "s3.amazonaws.com/book-covers.nypl.org",
    "book-covers.nypl.org",
    "covers.nypl.org"
]

target = 'https://covers.nypl.org'
class Replacement(object):
    def __init__(self, replace):
        self.replace = replace
        self.child = "mirror_url"

    def sub(self, o):
        self.child = Replacement(o)
        return self.child

    def __str__(self):
        return "replace(%s, '%s', '%s')" % (
            str(self.child), self.replace, target
        )


top = None
r = None
for prefix in prefixes:
    for suffix in replace:
        url = prefix + suffix
        if url == target:
            continue
        if not r:
            r = Replacement(url)
            top = r
        else:
            r = r.sub(url)

print "update representations set mirror_url=%s where mirror_url like '%%covers.nypl.org%%';" % str(top)
```

PR #182, opened 2018-08-24
Use the ErrorHandler from core

leonardr:
This branch makes the metadata wrangler use the same ErrorHandler class to handle exceptions that the circulation manager uses. Exceptions that affect database integrity will kill the process; other exceptions will be logged.

The motivating case is to make NYPL's metadata wrangler log the exceptions it encounters to Loggly so I don't have to SSH in and look at uwsgi.log.

This branch also gives some more specific error messages to the problem detail documents returned upon failures of the registration process.

Note that ErrorHandler has some code for rolling back a database session if there's an exception, which  won't work in the metadata wrangler because it doesn't define `app.manager`. We could certainly improve this, but it's not a big deal because the metadata wrangler has its own code for rolling back a database session if there's an exception.

### DIFF
--- a/app.py
+++ b/app.py
@@ -10,6 +10,7 @@ from flask_babel import Babel
 from flask_sqlalchemy_session import flask_scoped_session
 
 from core.app_server import (
+    ErrorHandler,
     HeartbeatController,
     returns_problem_detail,
 )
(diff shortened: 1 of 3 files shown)


app.py
```diff
@@ -59,6 +60,14 @@ def initialize_database(autoinitialize=True):
     for logger in logging.getLogger().handlers:
         app.log.info("Logs are going to %r", logger)
 
+    # Register an error handler that logs exceptions through the
+    # normal logging process and tries to turn them into Problem
+    # Detail Documents.
+    h = ErrorHandler(app, app.config['DEBUG'])
+    @app.errorhandler(Exception)
+    def exception_handler(exception):
+        return h.handle(exception)
```
aslagle: Is it necessary to put this in `app.before_first_request`? It probably doesn't matter, but I don't think we do that in the circ manager.

leonardr: My thinking is that `before_first_request` is the first place where we know the correct value for `app.config['DEBUG']`. It's not that way in the circ manager but it probably should be.

PR #185, opened 2018-11-08
Recognize and ignore empty cover images from Content Cafe

leonardr:
This branch fixes the Content Cafe part of https://jira.nypl.org/browse/SIMPLY-340.

I'm having trouble coming up with an effective migration, because most of the bad cover images on the metadata wrangler don't do any harm -- the books I spot-checked got their cover images from some other source, so the fact that there was a lousy cover associated with some distantly equivalent ISBN didn't hurt anything. So this problem may have only affected a couple of books, and it might be more effective to just come up with a one-off script and run it against the metadata wrangler.

At any rate, I want to get this code onto the metadata wrangler so that the problem doesn't become worse while I'm investigating it.


aslagle:
Robert had some RBdigital books with this problem but they were probably audiobooks.

aslagle:
What's the reason for comparing with the prefix instead of the whole image? The prefix image only cuts off a little bit of the bottom - it could probably be quite a bit smaller if that improves anything.

leonardr:
There's not just one image. Content Cafe sticks a little HTML page at the end of the cover image -- this is probably a bug on their side -- and the HTML page contains the ISBN. So every image looks the same but is a slightly different binary.

leonardr:
I bit the bullet and wrote a Python migration script that hunts down bad works and removes their cover images.

migration/20181108-remove-bad-content-cafe-covers.py
```diff
@@ -0,0 +1,47 @@
...
+        Representation.image_width==80).filter(
+            Representation.url.like("http://contentcafe2.btol.com/%")
+        )
+policy = PresentationCalculationPolicy.recalculate_everything()
+for rep in qu:
+    identifiers = [h.identifier for h in rep.resource.links]
+    fix_editions = []
+    for identifier in identifiers:
+        for edition in identifier.primarily_identifies:
+            if 'Content' in edition.cover_thumbnail_url or 'Content' in edition.cover_full_url:
```
leonardr: This check is necessary because sometimes the work ends up with a cover image from a more reliable source like Overdrive.

migration/20181108-remove-bad-content-cafe-covers.py
```diff
@@ -0,0 +1,47 @@
...
+    Representation,
+    get_one,
+    PresentationCalculationPolicy,
+)
+
+_db = production_session()
+qu = _db.query(Representation).filter(
+    Representation.image_height==120).filter(
+        Representation.image_width==80).filter(
+            Representation.url.like("http://contentcafe2.btol.com/%")
```
aslagle: Does this get all content cafe covers or only the placeholders? Are most of the covers different sizes?

leonardr: As far as I can tell it only gets the placeholders. The real covers are bigger than this.

PR #186, opened 2018-11-09
Move code for both OCLC APIs into a new package.

leonardr:
This is a cleanup branch in preparation for work on https://jira.nypl.org/browse/SIMPLY-1364. I just moved the code and test files that deal with OCLC APIs into their own packages.


oclc/classify.py
```diff
@@ -592,7 +592,11 @@ def lookup_by(self, **kwargs):
 
 
 class OCLCClassifyCoverageProvider(IdentifierCoverageProvider):
-    """Does title/author lookups using OCLC Classify."""
+    """Does title/author lookups using OCLC Classify.
+
+    NOTE: This code is no longer used. It was designed for Project Gutenberg,
+    where
```
aslagle: ?

leonardr: Oops, that's a bit of my next branch.

PR #187, opened 2018-11-09
Move out old code

leonardr:
Another simple cleanup branch in preparation for https://jira.nypl.org/browse/SIMPLY-1364.

Commented-out code was removed and OCLCClassifyCoverageProvider was renamed to TitleAuthorLookupCoverageProvider, since pretty soon there will be multiple CoverageProviders that use OCLC Classify.

### DIFF
--- a/coverage.py
+++ b/coverage.py
@@ -43,14 +43,6 @@ from content_cafe import (
     ContentCafeAPI,
 )
 
-from oclc.classify import (
-    OCLCClassifyCoverageProvider,
-)
-
-from oclc.linked_data import (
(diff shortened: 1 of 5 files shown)


aslagle:
Maybe I'm missing something but it looks like you didn't actually change the name of the class, just other references to it.

leonardr:
That's annoying. I must have lost the commit somehow.

leonardr:
OK, give this another look. It was a blessing in disguise that this happened because it revealed that the tests in _oclc weren't actually running.

PR #188, opened 2018-12-04
Look up books based on ISBN and parse the OCLC Classify responses

leonardr:


oclc/classify.py
```diff
@@ -26,15 +29,204 @@
...
+        contributors = cls.contributors(_db, tree)
+        measurements = cls.measurements(_db, tree)
+        subjects = cls.subjects(_db, tree)
+
+        return Metadata(
+            data_source=DataSource.OCLC,
+            contributors=contributors,
+            measurements=measurements,
+            subjects=subjects,
+        )
```
leonardr: Before adding tests there's one more change I'd like you to make.

I asked the OCLC Classify API about 73000 ISBNs and saved the XML files. We only got three different 'status codes':

2 (Single work detail): 78%
4 (Multi work detail): 10%
102 (not found): 11%

So those are the three the parser needs to handle.

Here's a bit of code from the old OCLCXMLParser for extracting the response code. I don't recommend copying any of the rest of the code from that class, except for the response code constants.

```
        response = cls._xpath1(tree, "oclc:response")
        representation_type = int(response.get('code'))
```

Your `parse` method should return a response code and a _list_ of Metadata objects. If the response code is 0 or 2, the list will contain one Metadata object -- that's basically the case you handle now. If the response code is 4, the list will contain several Metadata objects (you have some of this code now; I'll mention it later). Otherwise the list will be empty.

oclc/classify.py
```diff
@@ -26,15 +29,204 @@
...
+        work_tags = cls.get_work_tags(tree)
+        if work_tags is not None:
+            for work_tag in work_tags:
+                author_string = work_tag.get('author')
+                if author_string:
+                    author_lists.append(NameParser.parse_multiple(author_string))
+
+        # If there are multiple lists of authors (as a result of multiple <work> tags),
+        # use the shortest list.
+        return sorted(author_lists, key=len)[0]
```
leonardr: Here's an example of a multi-work response:

```
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<classify xmlns="http://classify.oclc.org">
  <response code="4"/>
  <!--Classify is a product of OCLC Online Computer Library Center: http://classify.oclc.org-->
  <workCount>2</workCount>
  <start>0</start>
  <maxRecs>25</maxRecs>
  <orderBy>thold desc</orderBy>
  <input type="isbn">9788888211756</input>
  <works>
    <work author="Manzoni, Alessandro, 1785-1873" editions="30" format="eAudiobook" holdings="59" hyr="2015" itemtype="itemtype-audiobook-digital" lyr="1900" owi="1151523604" schemes="DDC LCC" title="I promessi sposi" wi="123476150"/>
    <work author="Manzoni, Alessandro, 1785-1873" editions="1" format="Book" holdings="1" hyr="2009" itemtype="itemtype-book" lyr="2009" owi="1154035336" schemes="DDC" title="I promessi sposi : audiolibro integrale" wi="555641639"/>
  </works>
</classify>
```

Here's my thinking:

1. The #1 thing we need out of this is the author's display name, which in this case is "Alessandro Manzoni". The next most important things are the popularity and subject information.
2. To have a decent chance of knowing the author's display name (as opposed to guessing it), we need the author's VIAF ID or LOC ID.
3. We have some popularity information here but nothing else -- no VIAF ID or LOC ID, no subject information.
4. We can't get that information from this document, but we can get it by doing a _second_ lookup based on one of the OWI (OCLC Work IDs) contained in this document.
5. So the best thing for the parser to do is to assemble a list of two Metadata objects, one for each of the `<work>` tags.
6. The author information (obtained through `NameParser.parse_multiple`) can go into those Metadata objects, but the really important thing is the `IdentifierData` containing the OWI (type=Identifier.OCLC_WORK).

So that's the case where `parse` returns multiple results. If we get that going I think we'll be done with the parser part; then it'll just be a case of improving the IdentifierLookupCoverageProvider to handle all these cases.

oclc/classify.py
```diff
@@ -769,6 +951,75 @@ def lookup_by(self, **kwargs):
...
+            return self.failure(identifier, e.message)
+
+    def _single(self, _db, tree, identifier):
+        return [self.parser.parse(self._db, tree, [identifier])]
+
+    def _multiple(self, _db, owi_numbers, identifier):
+        results = []
+        for number in owi_numbers:
+            owi, is_new = Identifier.for_foreign_id(_db, Identifier.OCLC_WORK, number)
+            owi.weight = 1
```
leonardr: This needs to be an IdentifierData, not an Identifier. The database Identifier will be created during the `metadata.apply` call.

oclc/classify.py
```diff
@@ -769,6 +951,75 @@ def lookup_by(self, **kwargs):
...
+        return results
+
+    def _get_new_tree(self, owi):
+        url = 'http://classify.oclc.org/classify2/Classify?owi=' + owi.identifier
+        representation, cached = Representation.get(self._db, url)
+        return etree.fromstring(representation.content, parser=etree.XMLParser(recover=True))
+
+    def _apply(self, metadata, identifier):
+        edition = self.edition(identifier)
+        work = self.work(edition.primary_identifier)
```
leonardr: I don't think you need to create a Work here, but we might want to create one at the end of `process_item`. We don't need a Work for every OCLC Work ID that we looked up -- those are fine as Editions.

oclc/classify.py
```diff
@@ -769,6 +951,72 @@ def lookup_by(self, **kwargs):
...
+        """Ask OCLC Classify about a single ISBN and create an Edition
+        based on what it says.
+        """
+        # Perform a title/author lookup.
+
+        metadata_list = []
+
+        try:
+            xml = self.api.lookup_by(isbn=identifier.identifier)
+            tree, code, owi_numbers = self.parser.initial_look_up(self._db, xml)
```
leonardr: Let's try calling `owi_numbers` something like `substitutes`, and have `initial_look_up` return the IdentifierData objects rather than the raw numbers.

That way we can understand `_single` as the 'we found the data' method, and `_multiple` as the 'we didn't find the data, but we found some equivalent items' method.

I think we can also avoid dealing with the `code` here -- we can have a rule that if `substitutes` is `None` then the document contains the data we want (so we call _single), and otherwise the document is useless except for the list of substitutes (so we should call _multiple).

tests/oclc_/test_identifier_lookup_coverage_provider.py
```diff
@@ -0,0 +1,247 @@
...
+        result = provider.process_item(id)
+        eq_(etree.tostring(provider.called_with["tree"]), etree.tostring(provider._get_tree(self.SINGLE_ISBN)))
+        eq_(provider.called_with["identifier"], id)
+        eq_(result, id)
+
+    def test_process_item_multi(self):
+        # Testing that, when process_item finds out that a document's status code is 4,
+        # it calls _multiple, passes in the correct OWIs, and
+        # returns the original ISBN.  Uses mocked versions of _get_tree, initial_look_up,
+        # and _multiple.
```
leonardr: You also need to test that if `_multiple` returns a list of things, `self._apply()` is called on each one with the origina identifier.

leonardr: This doesn't need to go into `test_process_item_single` or `test_process_item_multi`; it can be a separate test to avoid clutter. We'll go over different strategies for doing this later, but I think the best way is to create a method like `MockProvider.queue` so that we can queue up the result we want, and change `_single` and `_multiple` to return the next thing off the queue.

tests/oclc_/test_identifier_lookup_coverage_provider.py
```diff
@@ -0,0 +1,247 @@
...
+
+class MockProviderSingle(MockProvider):
+    def _get_tree(self, isbn):
+        xml = sample_data("single_work_with_isbn.xml", "oclc_classify")
+        return etree.fromstring(xml, parser=etree.XMLParser(recover=True))
+
+class MockProviderMulti(MockProvider):
+    def _get_tree(self, isbn):
+        xml = sample_data("multi_work_with_owis.xml", "oclc_classify")
+        return etree.fromstring(xml, parser=etree.XMLParser(recover=True))
```
leonardr: The `isbn` argument isn't used here -- is it left over from when this worked differently?

tests/oclc_/test_identifier_lookup_coverage_provider.py
```diff
@@ -0,0 +1,247 @@
...
+        code, owi_data = provider.parser.initial_look_up(self._db, tree)
+        results = provider._multiple(self._db, owi_data, identifier)
+
+        # The document contained two <work> tags and therefore two OWIs, so we
+        # end up with two results.  They should both be Metadata objects, and
+        # should have the same data source and ISBN.
+
+        eq_(len(results), 2)
+        for result in results:
+            eq_(result.__class__.__name__, "Metadata")
```
leonardr: The built-in `isinstance` function will help you with this.

tests/oclc_/test_identifier_lookup_coverage_provider.py
```diff
@@ -0,0 +1,247 @@
...
+    def _get_subjects(self, subjects):
+        # Everything in the list of subjects should be a SubjectData object.
+        eq_(len([x for x in subjects if x.__class__.__name__ == "SubjectData"]), len(subjects))
+        # Extract a sublist for each type of classifier.
+        sublists = [[x for x in subjects if x.type == type] for type in ["DDC", "LCC", "FAST"]]
+        # There should always be 1 DDC classification and 1 LCC classification.
+        eq_((len(sublists[0]), len(sublists[1])), (1, 1))
+        return sublists
+
+    def _check_measurements(self, measurements, values):
```
leonardr: We did really thorough checks of the parsed data in the test of the parser itself, so I think spot checks are okay in these tests. That said, the code already exists, so it might be simpler to leave it alone.

tests/oclc_/test_classify_xml_parser.py
```diff
@@ -0,0 +1,103 @@
...
+        return sample_data(filename, 'oclc_classify')
+
+    def tree(self, filename):
+        xml = self.sample_data(filename)
+        return etree.fromstring(xml, parser=etree.XMLParser(recover=True))
+
+    def test_initial_look_up(self):
+        single_tree = self.tree("single_work_response.xml")
+        code, owi_data = self.parser.initial_look_up(self._db, single_tree)
+        eq_(2, code)
```
leonardr: It'll be more comprehensible if you use the OCLCClassifyXMLParser constants instead of these numbers.

tests/oclc_/test_classify_xml_parser.py
```diff
@@ -0,0 +1,103 @@
...
+        eq_(1, len(owi_data))
+        eq_(Identifier.OCLC_WORK, owi_data[0].type)
+        eq_(None, owi_data[0].identifier)
+        eq_(1, owi_data[0].weight)
```
leonardr: A lot of code like this can be simplified by doing `[owi] = owi_data`. That will crash unless there's exactly one item in `owi_data`, and then you have the thing you're testing in `owi` and you can work on it from that point.

coverage.py
```diff
@@ -242,16 +256,11 @@ def instantiate(cls, add_to, provider_kwargs, **kwargs):
...
-            )
-        except CannotLoadConfiguration, e:
-            # No Overdrive collection is configured -- we can't
-            # handle Overdrive lookups, as much as we'd like to.
-            pass
+        overdrive = instantiate(
+            OverdriveBibliographicCoverageProvider, providers,
+            provider_kwargs, collection=self.collection,
+            viaf=self.viaf, replacement_policy=self.replacement_policy
+        )
```
leonardr: This stuff is already in master; I don't know why it's showing up in the diff here.

oclc/classify.py
```diff
@@ -769,6 +954,72 @@ def lookup_by(self, **kwargs):
...
+        metadata_list = []
+
+        try:
+            tree = self._get_tree(isbn=identifier.identifier)
+            code, owi_data = self.parser.initial_look_up(self._db, tree)
+
+            if code in [0, 2]:
+                metadata_list = self._single(self._db, tree, identifier)
+            elif code == 4:
+                metadata_list = self._multiple(self._db, owi_data, identifier)
```
leonardr: Again, using the previously defined constants will make it obvious what's going on here. Also, you need to handle the NOT_FOUND_STATUS by returning a transient `CoverageFailure` object. [See here](https://github.com/NYPL-Simplified/server_core/blob/672f4ede9c3d1ac7b369291fa77b179a7acf0053/opds_import.py#L581) for a similar example. (That one is inside an exception handler, so it has a traceback to generate the failure message -- you would write the failure message yourself.)

oclc/classify.py
```diff
@@ -769,6 +955,72 @@ def lookup_by(self, **kwargs):
...
+
+        try:
+            tree = self._get_tree(isbn=identifier.identifier)
+            code, owi_data = self.parser.initial_look_up(self._db, tree)
+            if code in [self.parser.SINGLE_WORK_DETAIL_STATUS, self.parser.SINGLE_WORK_SUMMARY_STATUS]:
+                metadata_list = self._single(self._db, tree, identifier)
+            elif code == self.parser.MULTI_WORK_STATUS:
+                metadata_list = self._multiple(self._db, owi_data, identifier)
+            elif code == self.parser.NOT_FOUND_STATUS:
+                return CoverageFailure(identifier, traceback.format_exc(), data_source=DataSource.OCLC, transient=True)
```
leonardr: This isn't quite right. You're not catching an exception here, so traceback.format_exc() won't do anything. Instead of importing `traceback`, just pass in an error message like "OCLC Classify reported not found".

oclc/classify.py
```diff
@@ -769,6 +955,72 @@ def lookup_by(self, **kwargs):
...
+            tree_from_owi = self._get_tree(owi=item.identifier)
+            results.append(self.parser.parse(self._db, tree_from_owi, [identifier, item]))
+
+        return results
+
+    def _apply(self, metadata, identifier):
+        edition = self.edition(identifier)
+        work = self.work(edition.primary_identifier)
+        metadata.apply(
+            edition, collection=None #, replace=self.replacement_policy
```
leonardr: You can take out the commented part of this line; it turns out we don't need it, I'm pretty sure.

oclc/classify.py
```diff
@@ -795,10 +1047,10 @@ class TitleAuthorLookupCoverageProvider(IdentifierCoverageProvider):
     SERVICE_NAME = "OCLC Classify Coverage Provider"
     INPUT_IDENTIFIER_TYPES = [Identifier.GUTENBERG_ID, Identifier.URI]
     DATA_SOURCE_NAME = DataSource.OCLC
-    
+
     def __init__(self, _db, api=None, **kwargs):
         super(TitleAuthorLookupCoverageProvider, self).__init__(
-            _db, registered_only=True, **kwargs
+            _db, **kwargs
```
leonardr: `registered_only=True` still needs to be present. Otherwise we'll go around trying to do a title/author lookup on identifiers that aren't actually being used.

oneoff_isbn_lookup.py
```diff
@@ -0,0 +1,19 @@
...
+    Identifier,
+)
+
+_db = production_session()
+
+isbn = sys.argv[1]
+identifier, is_new = Identifier.for_foreign_id(_db, Identifier.ISBN, isbn)
+collection, is_new = Collection.by_name_and_protocol(_db, name="a", protocol="b")
+provider = IdentifierLookupCoverageProvider(collection=collection)
+provider.process_item(identifier)
```
leonardr: Remove the oneoff script to make the branch ready for review.

tests/oclc_/test_identifier_lookup_coverage_provider.py
leonardr: I figured this out. https://github.com/NYPL-Simplified/metadata_wrangler/pull/188/files#diff-5970034b7fe3730fa953b77d85b6de74R976 defines `_get_tree()` to take `kwargs`, with the expectation that either an `isbn` or a `owi` will be passed in. So I recommend defining your mock `_get_tree()` implementations to take `kwargs` and ignore them. That way there's no mystery about why a specific named argument is being ignored.

tests/oclc_/test_title_author_lookup_classify.py
```diff
@@ -16,7 +16,7 @@
 )
 
 from oclc.classify import (
-    OCLCXMLParser,
```
leonardr: Note to Amy: this file is just reflecting the new name of the OCLCTitleAuthorLookupXMLParser.

tests/test_coverage.py
```diff
@@ -22,40 +22,28 @@
 from viaf import VIAFClient
 
 
+class MockProvider(IdentifierResolutionCoverageProvider):
```
leonardr: Again, the changes to this file are already in master, I don't know why they're showing up in this diff.

coverage.py
```diff
@@ -126,7 +126,14 @@ def __init__(self, collection, mirror=None, http_get=None, viaf=None,
```
leonardr: The changes to this file are already in master; I don't know why they're showing up in this diff.

aslagle: Merging master into the branch would probably fix it.

oclc/classify.py
```diff
@@ -26,15 +30,284 @@
...
+
+
+    @classmethod
+    def _parse_subject_tag(cls, tag):
+        """Extract from an etree tag the information necessary to
+        create a SubjectData.
+        """
+        # The names of the attributes vary depending on which type of classifier
+        # the tag is for.  Only Fast tags have a name.
+        weight = tag.get("holdings") or tag.get("heldby")
```
aslagle: I'm not sure how the weight is used but it seems like it would need to be normalized to be useful. Would it make sense to divide by the total holdings for the work?

leonardr: I hadn't thought about this, but having thought about it, I think leaving it as-is is the best solution.

The weights measure the strength of the evidence. We grab all the Classifications for an Identifier (and other Identifiers equivalent to that one), put them all in a weighted data structure, and see which genres are the most strongly implied. So if 1000 libraries classify a book as fiction and 10 as nonfiction, then fiction wins.

If there's only one Identifier (which is going to be the case most of the time, until we bring OCLC Linked Data back online) then the absolute numbers don't matter. 1000 vs. 10 is the same as 1 versus 0.001.

But then let's say we bring in data from another source like OCLC Linked Data. The Linked Data subjects aren't weighted, so they all end up with a weight of 1. So for most ISBNs, the classifications from OCLC Classify are going to swamp the classifications from OCLC Linked Data.

In one sense that's not fair to OCLC Linked Data. But knowing that 1000 libraries classified a book a certain way is really strong evidence. OCLC Linked Data doesn't provide that evidence, so it can only help on the margins, or when OCLC Classify data is unavailable.

This is a big part of the reason I disabled the OCLC Linked Data coverage provider -- it was doing a lot of work little benefit. I'm hoping that OCLC Classify + VIAF will give us almost all the benefits of the old system at a much lower cost.

PR #189, opened 2018-12-05
Add standalone, testable parsers for OCLC Classify and VIAF's name formats

leonardr:
This is self-contained work in service of https://jira.nypl.org/browse/SIMPLY-1364.

This branch adds two classes called NameParser which parse two very similar formats for the names of human beings. They do not try to parse the names themselves, e.g. convert sort_name to display_name. They just try to convert a person's name into a ContributorData object, possibly by extracting information that's _not_ part of a person's name and putting it in some other field of ContributorData.

[VIAF names](http://viaf.org/viaf/112722880/viaf.xml) look like "Dibdin, Michael, 1947-2007." I don't know if these names were _originated_ by VIAF but it's a very common style. Here the potential extra data is the birth and death dates.

[OCLC names](http://classify.oclc.org/classify2/Classify?isbn=9780316509695&summary=false) are based on VIAF names (and so, OCLC's NameParser subclasses VIAF's NameParser). But since OCLC names are given in the context of a specific work, they often have roles associated with them. e.g. "Patterson, Jack, 1998- [Author]"

In addition, OCLC names are sometimes given in a pipe-separated list, like this: "Patterson, James, 1947- | Madsen, Jim, 1964- [Illustrator] | Patterson, Jack, 1998- [Author] | Yue, Florence [Art director]". OCLC's NameParser can also parse this list.


oclc/classify.py
```diff
@@ -32,6 +36,147 @@ class OCLC(object):
...
+     Leodhas, Sorche Nic, 1898-1969 | Ness, Evaline [Illustrator]
+    """
+
+    ROLES = re.compile("\[([^]]+)\]$")
+
+    # Map the roles defined in OCLC Classify to the constants
+    # defined in Contributor.
+    ROLE_MAPPING = {
+        "Author": Contributor.AUTHOR_ROLE,
+	"Translator": Contributor.TRANSLATOR_ROLE,
```
aslagle: I think you have a mixture of spaces and tabs here and it's messing up the indentation in github.

oclc/classify.py
```diff
@@ -32,6 +36,147 @@ class OCLC(object):
...
+                    # explicitly specified. If we see someone with no
+                    # explicit role after this point, it's probably
+                    # because their role is so minor as to not be
+                    # worth mentioning, not because it's so major that
+                    # we can assume they're an author.
+                    default_role = Contributor.UNKNOWN_ROLE
+            else:
+                # No explicit role was provided. Assign the default
+                # role.
+                contributor.roles = [default_role]
```
aslagle: If we assigned someone the primary author role at this point, should we change the default role to author here too?

leonardr: Yeah, I think we might need to make a little state machine here because the same logic applies when we're parsing multiple authors from subsequent <author> tags.

leonardr:
@aslagle I think this is ready now.

oclc/classify.py
```diff
@@ -103,28 +103,62 @@ def parse_multiple(cls, author_string):
...
-                    # explicitly specified. If we see someone with no
-                    # explicit role after this point, it's probably
-                    # because their role is so minor as to not be
-                    # worth mentioning, not because it's so major that
-                    # we can assume they're an author.
-                    default_role = Contributor.UNKNOWN_ROLE
-            else:
-                # No explicit role was provided. Assign the default
-                # role.
-                contributor.roles = [default_role]
```
leonardr: It wasn't necessary to set .roles -- that happens in `parse`.

viaf.py
```diff
@@ -20,23 +20,82 @@
...
+
+    These are used by VIAF but also (in slightly modified form) by
+    OCLC Classify.
+
+    TODO: VIAFParser has its own "name parsing" code which focuses on
+    extractingdata from the XML generated by the VIAF API. These code
+    bases need to be merged.
+    """
+
+    LIFESPAN = re.compile("([0-9]*)?\??-([0-9]*)\???[.;]?$")
```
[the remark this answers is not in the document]
aslagle: Oh I see - it's because the captured group already has `*` in it, so it can be empty even without a question mark after it.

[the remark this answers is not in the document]
aslagle: You may have more question marks than necessary here, but I think it looks fine.

PR #192, opened 2018-12-21
Build a single Metadata object instead of building a number of them and applying them to the same Edition.

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-1583. Adriana has fixed the tests and I've verified that it works in a real environment.

In the multi-work case, we were making a Metadata object for each OWI, but instead of giving each Metadata object an OWI as its primary_identifier, we were giving it the original ISBN. When we called apply() on the first Metadata object, it associated all of its ContributorData, SubjectData, and MeasurementData with the original ISBN. That's fine, but when we called apply() on the second Metadata object:

* It associated all of its ContributorData with the original ISBN, adding to the first list of ContributorData. This had the effect of listing the same author multiple times.
* It overwrote the first set of Measurements for that ISBN with the second set. So if OWI #1 had 10 editions and OWI #2 had 13 editions, the final Measurement would be 13, not 23.
* It overwrote the first set of Classifications for that ISBN with the second set. So if OWI #1 had a weight of 102 for a certain subject, and OWI #2 had a weight of 10, the final Classification would have a weight of 10, not 112.

This branch changes the architecture so that we start with a single Metadata object, for the ISBN, and we improve it gradually as we iterate over the OWIs. This involves some code for detecting whether a MeasurementData already exists for a certain quantity (in which case we bump up the value measured) and whether a SubjectData already exists for a certain subject (in which case we bump up the weight).

This branch also addresses another problem I found while doing QA on the original OCLC Classify branch. Quantities are being stored under OCLC's names for them: 'holdings' and 'editions'. The names we should be using are constants defined in the Measurement class: HOLDINGS and PUBLISHED_EDITIONS.

### DIFF
--- a/oclc/classify.py
+++ b/oclc/classify.py
@@ -157,7 +157,7 @@ class OCLCClassifyXMLParser(XMLParser):
         and update a list of existing MeasurementData.
         """
         tags = cls._xpath(tree, "//oclc:work")
-        for measurement in cls.get_measurements(tags, existing_measurements):
+        for measurement in cls.get_measurements(tags):
             key = measurement.quantity_measured
             if key in existing_measurements:
                 # We don't need another MeasurementData -- just add it to
@@ -177,7 +177,7 @@ class OCLCClassifyXMLParser(XMLParser):
     }
 
     @classmethod
-    def get_measurements(cls, work_tags, existing_measurements):
+    def get_measurements(cls, work_tags):
         """Extract MeasurementData representing values
         measured from a list of <work> tags.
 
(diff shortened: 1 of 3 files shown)


aslagle:
It looks like core went backwards here but I'm not sure if it matters.

oclc/classify.py
```diff
@@ -144,16 +152,32 @@ def _add_lc_viaf(cls, author, tag):
...
+    # creating Measurements.
+    MEASUREMENT_MAPPING = {
+        "holdings" : Measurement.HOLDINGS,
+        "eholdings" : Measurement.HOLDINGS,
+        "editions" : Measurement.PUBLISHED_EDITIONS,
+    }
 
     @classmethod
-    def get_measurements(cls, work_tags):
+    def get_measurements(cls, work_tags, existing_measurements):
```
aslagle: It looks like this function doesn't need the existing measurements.

PR #193, opened 2019-01-04
Use the same ReplacementPolicy for all metadata coverage providers

leonardr:
This is a follow-up branch to fix https://jira.nypl.org/browse/SIMPLY-1583. It defines MetadataWranglerBibliographicCoverageProvider._default_replacement_policy to provide a reasonable default, so that individual coverage providers don't have to individually define a ReplacementPolicy. The problem remaining in 1583 was that the OCLC Classify coverage provider was using a ReplacementPolicy that added to lists of contributors instead of replacing contributors, so attempts to fix the problem only made it worse.

I spent quite a while tracking down a problem that was caused by an error in a migration script. Basically, if an ISBN was in multiple collections, we were registering multiple CoverageRecords for it, and duplicate work was being done in an extremely inefficient way. This branch includes a lot of defensive code so that if this happens again, the duplicate work will still happen, but it'll be much more efficient and reliable.

I've fixed the duplicate CoverageRecords on the metadata wrangler, and I've modified the migration script to say what it should have said in the first place. I've left the defensive code in place, even though it should no longer happen in a real environment.

### DIFF
--- a/coverage_utils.py
+++ b/coverage_utils.py
@@ -1,14 +1,36 @@
 from nose.tools import set_trace
+
+from sqlalchemy.orm.session import Session
+
+from core.config import CannotLoadConfiguration
 from core.coverage import (
     BibliographicCoverageProvider,
     CoverageFailure
 )
+from core.metadata_layer import ReplacementPolicy
+from core.mirror import MirrorUploader
 from core.model import (
     DataSource,
     Work,
 )
 
 class MetadataWranglerBibliographicCoverageProvider(BibliographicCoverageProvider):
+
+    def _default_replacement_policy(self, _db):
+        """In general, data used by the metadata wrangler is a reliable source
+        of metadata but not of licensing information. We always
+        provide the MirrorUploader in case a data source has cover
+        images available.
+        """
+        try:
+            mirror = MirrorUploader.sitewide(_db)
+        except CannotLoadConfiguration, e:
+            # It's not a fatal error if there's no MirrorUploader
+            # configured -- it just means we can't mirror cover images
+            # when they show up.
+            self.log.error("No sitewide uploader configured", exc_info=e)
+            mirror = None
+        return ReplacementPolicy.from_metadata_source(mirror=mirror)
 
     def work(self, identifier):
         """Create or find a Work for the given Identifier.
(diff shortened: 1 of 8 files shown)


coverage_utils.py
```diff
@@ -17,17 +46,61 @@ def work(self, identifier):
...
+        # COVERAGE_COUNTS_FOR_EVERY_COLLECTION. But migration scripts
+        # may register seemingly redundant work to be done, and if
+        # that happens, we don't need to create a whole other Work
+        # -- we just need to recalculcate its presentation, which will
+        # happen in handle_success().
+        existing_work = identifier.work
+        if existing_work:
+            license_pool.work = existing_work
+            return existing_work
+
```
leonardr: I don't have a test for this because it's purely a performance improvement.

coverage_utils.py
```diff
@@ -17,17 +46,61 @@ def work(self, identifier):
...
+        """
+        work = self.work(identifier)
+        if not isinstance(work, Work):
+            return work
+
+        if work.presentation_ready:
+            # This work was already presentation-ready, which means
+            # its presentation probably just changed and needs to be
+            # recalculated.
+            work.calculate_presentation()
```
leonardr: This may not technically be necessary because Metadata.apply will flag the work as needing its presentation recalculated. I like having it here because it lets me go to a URL like https://metadata.librarysimplified.org/lookup?urn=urn:isbn:9780062019622&resolve_now=true and see the new data immediately.

migration/20181220-register-isbns-for-oclc-coverage.sql
```diff
@@ -4,13 +4,15 @@ delete from coveragerecords where id in (
...
--- record in the 'registered' state for the OCLC CLassify coverage
--- provider.
+-- For every ISBN associated with a collection and not already
+-- registered, create a coverage record in the 'registered' state for
+-- the OCLC CLassify coverage provider.
 insert into coveragerecords (
  identifier_id, data_source_id, timestamp, status
 ) select 
- i.id, ds.id, now(), 'registered'
+ distinct(i.id), ds.id, now(), coverage_status('registered')
```
leonardr: The distinct() is what was missing from the original script. The rest is something I added to make it safer to add coverage records without creating duplicates.

PR #194, opened 2019-02-05
Don't treat the string 'No content currenly exists for this item' as the title of a book.

leonardr:
This fixes https://jira.nypl.org/browse/SIMPLY-1700

### DIFF
--- a/migration/20190205-remove-content-cafe-coverage-for-books-with-no-title.sql
+++ b/migration/20190205-remove-content-cafe-coverage-for-books-with-no-title.sql
@@ -1,0 +1,11 @@
+-- Restore all matching books to the 'registered' state so Content
+-- Cafe will process them again.
+update coveragerecords set status='registered' where identifier_id in (select primary_identifier_id from editions where title='No content currently exists for this item') and data_source_id in (select id from datasources where name='Content Cafe') and operation is null;
+
+-- Remove the records of whatever the Content Cafe coverage provider
+-- did the first time.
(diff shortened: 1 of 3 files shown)


leonardr:
I've verified that there's no other obviously non-title string that shows up a lot as a title in the metadata wrangler.

migration/20190205-remove-content-cafe-coverage-for-books-with-no-title.sql
```diff
@@ -0,0 +1,11 @@
...
+-- Cafe will process them again.
+update coveragerecords set status='registered' where identifier_id in (select primary_identifier_id from editions where title='No content currently exists for this item') and data_source_id in (select id from datasources where name='Content Cafe') and operation is null;
+
+-- Remove the records of whatever the Content Cafe coverage provider
+-- did the first time.
+delete from coveragerecords where identifier_id in (select primary_identifier_id from editions where title='No content currently exists for this item') and data_source_id in (select id from datasources where name='Content Cafe') and operation is not null;
+
+-- Remove the bad data manually. The new 'title' is null, and the
+-- metadata layer will not overwrite a non-null value with null.
+update editions set title=null where title='No content currently exists for this item';
```
EdwinGuzman: Python's `None` is equivalent to sql's `null`?

leonardr: Yep.

PR #195, opened 2019-03-26
Make the canonicalization controller able to handle invalid identifier URIs.

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-1816, which causes the 500 error you can see by going [here](http://metadata.librarysimplified.org/canonical-author-name?display_name=Jeremy%20Robinson&urn=urn%3Aisbn%3ADBKACX0122823).

Passing an invalid URN into `Identifier.parse_urn` will raise a `ValueError`. Most of the time this is correct--a bad URN should put a stop to whatever we're trying to do--but in the context of the canonicalization controller, the URN is optional. If a bad URN is passed in, we should try to find the author's canonical name based solely on their display name.

Since the `CanonicalizationController` had no test coverage, I added tests for the whole thing.

### DIFF
--- a/canonicalize.py
+++ b/canonicalize.py
@@ -263,6 +263,10 @@ class AuthorNameCanonicalizer(object):
 
 class MockAuthorNameCanonicalizer(AuthorNameCanonicalizer):
 
+    """Mocks the services used by the author name canonicalizer, but
+    leaves the logic alone.
+    """
+
     def __init__(self, _db, oclcld=None, viaf=None):
         super(MockAuthorNameCanonicalizer, self).__init__(_db)
         self._db = _db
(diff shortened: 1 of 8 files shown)


canonicalize.py
```diff
@@ -263,6 +263,10 @@ def sort_name_from_viaf(self, display_name, known_titles=None):
 
 class MockAuthorNameCanonicalizer(AuthorNameCanonicalizer):
 
+    """Mocks the services used by the author name canonicalizer, but
+    leaves the logic alone.
+    """
+
```
leonardr: We don't want to use this mock service when testing the controller because you still have to orchestrate the OCLC Linked Data and VIAF APIs. We want to mock the `canonicalize_author_name` method itself, so I made a new mock class.

I don't think this actually needs to be a mock class -- you should be able to pass mock OCLC Linked Data and VIAF APIs into the regular constructor.

leonardr:
The test failures are probably not related to this branch; I'm looking into them.

leonardr:
The test failures happened because some of the unit tests for OCLC Classify were making real HTTP requests to the API, asking about a specific book, and parsing the live XML sent from the API. Between the last time we changed this branch and now, OCLC's data about that book (e.g. the number of libraries that have the book in their holdings) changed slightly, and our tests started failing because we were verifying that the parser got the old answers when it finished its work.

Of course, we should not be making real HTTP requests during unit tests. I added a mock OCLCClassifyAPI and changed all the tests that were making real HTTP requests to use it. I made local copies of the files we were retrieving from OCLC Classify and ran the entire test suite on my laptop in airplane mode to verify that no other live HTTP requests were bring made.

Since we don't have the old data about number of holdings, I updated the tests to use the new values. We won't need to update it again.

PR #196, opened 2019-03-28
Detect and handle the case where we're asked to canonicalize a 'display name' that's actually a sort name.

leonardr:
With very little work, this will prevent almost all of the metadata wrangler timeouts described in https://jira.nypl.org/browse/SIMPLY-1818.

### DIFF
--- a/tests/test_canonicalize.py
+++ b/tests/test_canonicalize.py
@@ -85,6 +85,26 @@ class TestAuthorNameCanonicalizer(DatabaseTest):
         eq_("Rand, Ayn", m('Rand, Ayn, and Cher'))
         eq_("Rand, Ayn", m('Rand, Ayn, and Kaling, Mindy'))
 
+    def test__canonicalize_single_name(self):
+        # For single-named entities, the sort name and display name
+        # are identical. We don't need to ask VIAF.
+        self.canonicalizer.viaf.queue_lookup("bad data")
+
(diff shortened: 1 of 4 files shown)


tests/test_canonicalize.py
```diff
@@ -48,11 +48,31 @@ def queue_file_in_mock_http(self, filename):
...
-        # Make sure only the first human's name is used.
-        name = "Mindy Kaling, Bob Saget and Co"
-        extracted_name = self.canonicalizer.primary_author_name(name)
-        eq_("Mindy Kaling", extracted_name)
+        # Test our ability to turn a freeform string that identifies
+        # one or more people into the likely name of one person.
+        m = self.canonicalizer.primary_author_name
+
+        # Test the simplest case.
+        eq_("Mindy Kaling", "Mindy Kaling")
```
aslagle: You forgot to call `m` here :)

aslagle:
Looks good once you fix the tests.

leonardr:
Hmm, okay, we have some problems caused by the submodule update. I'll look into that.

I looked through the 500 errors we were getting and noticed that "Various" causes a 500 error not because of this problem, but because asking VIAF about 'Various' is a big mistake. It's also not necessary -- for entities with just one name, the display name and sort name are identical. I'll fix that as well.

PR #197, opened 2019-08-13
Use the zeep SOAP client instead of suds.

leonardr:
This branch resolves https://jira.nypl.org/browse/SIMPLY-2210. It removes the only external dependency in the Library Simplified system that's not available in Python 3: the suds SOAP client.

This was pretty easy. We only use SOAP for one thing, and the zeep client is very similar to suds, to the point that I think it might be a successor or a fork. However, we didn't have test coverage of our SOAP client usage, and I had to add it to make sure that I didn't break anything. And there are, in fact a couple of small differences regarding how the SOAP client serializes missing or empty values.

Since this is the first metadata branch in a while, there are also some changes to reflect recent changes in Travis CI and the core submodule.


content_cafe.py
```diff
@@ -354,48 +351,49 @@ class ContentCafeSOAPError(IOError):
...
+        self.soap = SOAPClient(wsdl_url, **kwargs)
 
     def get_content(self, key, content):
         data = self.soap.service.Single(
             userID=self.user_id, password=self.password,
-            key=key, content=content)
-        if hasattr(data, 'Error'):
+            key=key, content=content
+        )
+        if data.Error:
```
leonardr: In sudz, if there was no error then `Error` wasn't an attribute of the serialized object. In zeep, it's always there (because it's defined by the schema) but it's None.

content_cafe.py
```diff
@@ -354,48 +351,49 @@ class ContentCafeSOAPError(IOError):
...
             raise ContentCafeSOAPError(data.Error)
         else:
             return data
 
-    def estimated_popularity(self, key, cutoff=None):
+    def estimated_popularity(self, key, cutoff=None, as_of=None):
```
leonardr: The popularity estimation algorithm depends on the current date, which obviously can't be used in a stable test. This lets us test what value the algorithm _would_ give as of different dates.

content_cafe.py
```diff
@@ -354,48 +351,49 @@ class ContentCafeSOAPError(IOError):
...
         gathered = self.gather_popularity(data)
-        return self.estimate_popularity(gathered, cutoff)
+        return self.estimate_popularity(gathered, cutoff, as_of)
 
     def gather_popularity(self, detail):
         by_year_and_month = Counter()
         [request_item] = detail.RequestItems.RequestItem
         items = request_item.DemandHistoryItems
-        if items == '':
+        if not items:
```
leonardr: Here, the difference is that sudz treated an empty list of DemandHistoryItems as the empty string (which seems weird in retrospect), and zeep treats it as `None`.

controller.py
```diff
@@ -311,7 +311,9 @@ def updates_feed(self, collection_details):
         precomposed_entries = []
         # Add entries for Works associated with the collection's catalog.
         updated_works = collection.works_updated_since(self._db, last_update_time)
-        works = pagination.apply(updated_works).all()
+        works = pagination.modify_database_query(
+            self._db, updated_works
+        ).all()
```
leonardr: This reflects a change made in core during the Elasticsearch work. `Pagination` can now  modify either a database query or an Elasticsearch query, which made `apply` too generic a name.

tests/test_content_cafe.py
```diff
@@ -453,3 +462,117 @@ def create_metadata(self, *args, **kwargs):
...
+
+    def post(self, address, data, *args, **kwargs):
+        return super(MockSession, self).do_post(address, data)
+
+
+class TestContentCafeSOAPClient(ContentCafeAPITest):
+
+    def setup(self):
+        wsdl = self.data_file("service.wsdl")
+        self.session = MockSession(wsdl)
```
EdwinGuzman: I don't know much about SOAP or wsdl but they are different XML protocols, right? How do they work together in this case?

tests/test_content_cafe.py
```diff
@@ -453,3 +462,117 @@ def create_metadata(self, *args, **kwargs):
...
+        # popularity past our threshold of relevancy -- which, let's
+        # say, is three months.
+        queue()
+        six_months_later = data_gather_date + datetime.timedelta(days=180)
+        three_months = datetime.timedelta(days=90)
+
+        # The book's popularity is now given at half of its all-time
+        # greatest popularity.  This reflects the fact that it used to
+        # be very popular, and is probably still a book that library
+        # patrons
```
EdwinGuzman: Not a big deal but I think it's missing some final thoughts.

EdwinGuzman:
Is ContentCafe the only client that uses SOAP?

tests/test_content_cafe.py
leonardr: SOAP is an HTTP protocol for making RPC calls. WSDL is an XML file format that defines the RPC calls and the format of the data that's returned. The WSDL file is how we know how to invoke an RPC method called "DemandHistoryItems". We know that once we call that method we'll get back a data structure that we can navigate as "RequestItems.RequestItem.DemandHistoryItems.DemandHistoryItem", and that each DemandHistoryItem has a Year and a Month. This is all stuff that Content Cafe made up and put into the WSDL file.

So a SOAP client is instantiated with a WSDL file, and that controls which functions you can call on the client.

leonardr:
> Is ContentCafe the only client that uses SOAP?

Yes, and only to get historical demand information.

PR #198, opened 2019-08-14
Make the Python 2 code forward-compatible with Python 3

leonardr:
All this stuff should be familiar from our work on core and circulation.

I'm less confident in the Python 3 conversion for this product because our test coverage isn't as good here as for circulation, core, and library_registry.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -1,3 +1,4 @@
+import base64 as stdlib_base64
 from nose.tools import set_trace
 from datetime import datetime
 from flask import request, make_response
@@ -11,7 +12,6 @@ from sqlalchemy.orm import joinedload
 from sqlalchemy.sql.expression import and_
 from Crypto.PublicKey import RSA
 from Crypto.Cipher import PKCS1_OAEP
-import base64
 import feedparser
 import json
 import jwt
@@ -63,6 +63,7 @@ from core.util.authentication_for_opds import AuthenticationForOPDSDocument
 from core.util.http import HTTP
 from core.util.opds_writer import OPDSMessage
 from core.util.problem_detail import ProblemDetail
+from core.util.string_helpers import base64
 
 from coverage import (
     IdentifierResolutionCoverageProvider,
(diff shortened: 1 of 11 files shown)


controller.py
```diff
@@ -63,6 +63,7 @@
 from core.util.http import HTTP
 from core.util.opds_writer import OPDSMessage
 from core.util.problem_detail import ProblemDetail
+from core.util.string_helpers import base64
```
leonardr: We use `base64` in a part of the file that isn't mentioned in this diff. The part that shows up in the diff is where we still need to use stdlib's base64, which has been renamed.

oclc/linked_data.py
```diff
@@ -64,25 +64,25 @@ def for_type(self, g, search):
 
     @classmethod
     def restrict_to_language(self, values, code_2):
-        if isinstance(values, basestring) or isinstance(values, dict):
+        if isinstance(values, (unicode, dict)):
             values = [values]
         for v in values:
-            if isinstance(v, basestring):
+            if isinstance(v, unicode):
```
leonardr: Any reference to `basestring` is problematic because Python 3 doesn't have the concept. In this file, we're only dealing with strings that were loaded from JSON, so I think it's safe to assume we're dealing with Unicode strings.

(Also, we don't use this code, it's really old and it's starting to rot -- not happy with this file in general.)

oclc/linked_data.py
```diff
@@ -187,7 +187,9 @@ def lookup(self, identifier_or_uri, processed_uris=set()):
         """Perform an OCLC Open Data lookup for the given identifier."""
         type = None
         identifier = None
-        if isinstance(identifier_or_uri, basestring):
+        if isinstance(identifier_or_uri, bytes):
+            identifier_or_uri = identifier_or_uri.decode("utf8")
+        if isinstance(identifier_or_uri, unicode):
```
leonardr: Here I do want to keep open the possibility that we're looking at a bytestring, because URIs for lookups can come in from outside.

controller.py
EdwinGuzman: But is this until the conversion or will this remain?

leonardr: I think we'll keep this, because it makes it easy to base64-encode a Unicode string in Python 3 and get a Unicode string back.

PR #199, opened 2019-08-16
Refactor controller code, fix some edge-case bugs, and start testing the routing code.

leonardr:
This branch refactors the controller code and the app setup code so that it's easy to test most of it. While doing this I found a number of bugs, but they were all pretty small and not likely to happen in a real situation.

The controllers are now grouped together under a `MetadataWrangler` object rather than being instantiated in routing code. This allows us to test the routing code using a framework adapated from circulation/tests/test_routes.py. We can't use exactly the same code because the `MetadataWrangler` works differently from the `CirculationManager` (it's a lot simpler).

I renamed `coverage.py` to `coverage_provider.py` so that I could run the code coverage tool called "coverage.py" on the source code.


leonardr:
Here's the report from coverage.py. I don't have a total coverage number because I couldn't get coverage.py to exclude the "core" and "tests" packages. (This isn't a problem for circulation because almost all the code is in its own package, "api".)

```
Name                                                      Stmts   Miss  Cover
-----------------------------------------------------------------------------
app.py                                                       73     46    37%
canonicalize.py                                             164     96    41%
content_cafe.py                                             217      6    97%
controller.py                                               419      4    99%
coverage_provider.py                                         91      1    99%
coverage_utils.py                                            55      2    96%
fast.py                                                      47     47     0%
integration_client.py                                        77      5    94%
monitor.py                                                   78     78     0%
oclc/__init__.py                                              0      0   100%
oclc/classify.py                                            573     72    87%
oclc/linked_data.py                                         647    204    68%
overdrive.py                                                 38      3    92%
problem_details.py                                            6      0   100%
routes.py                                                    56      0   100%
scripts.py                                                  371    371     0%
shadow_catalog.py                                           145    145     0%
viaf.py                                                     432     51    88%
```

controller.py
```diff
@@ -81,46 +83,157 @@
...
-    if header and 'bearer' in header.lower():
-        shared_secret = base64.b64decode(header.split(' ')[1])
-        client = IntegrationClient.authenticate(_db, shared_secret)
-        if client:
-            return client
-    if not required and not header:
-        # In the case that authentication is not required
-        # (i.e. URN lookup) return None instead of an error.
-        return None
-    return INVALID_CREDENTIALS
```
leonardr: This is implemented in a more robust form as a class method of MetadataWrangler. That makes it easy to swap in a mocked implementation when we're testing the @accepts_auth and @requires_auth annotators.

controller.py
```diff
@@ -81,46 +83,157 @@
...
+        self.catalog = CatalogController(self._db)
+        self.integration = IntegrationClientController(self._db)
 
-def collection_from_details(_db, client, collection_details):
-    if not (client and isinstance(client, IntegrationClient)):
-        return None
+    @classmethod
+    def authenticated_client_from_request(cls, _db, required=True):
+        """Look up the IntegrationClient that's making the active
+        request.
```
leonardr: This is a class method because that makes it easy to call in different situations (testing vs not, decorator vs inside a controller).

controller.py
```diff
@@ -81,46 +83,157 @@
...
+            client = IntegrationClient.authenticate(_db, shared_secret)
+
+            if client:
+                # Success!
+                flask.request.authenticated_client = client
+                return client
+            else:
+                # The bearer token doesn't identify any particular
+                # IntegrationClient.
+                return INVALID_CREDENTIALS
```
leonardr: Previously, if you passed in bad credentials to do something that allowed but didn't require authentication, you would perform the action as though you had not authenticated. I think it's better if your bad credentials are rejected; that way you know there's a problem.

controller.py
```diff
@@ -81,46 +83,157 @@
...
+        self.heartbeat = HeartbeatController()
+        self.canonicalization =  CanonicalizationController(self._db)
+        self.index = IndexController(self._db)
+        self.urn_lookup = URNLookupController(self._db)
+        self.catalog = CatalogController(self._db)
+        self.integration = IntegrationClientController(self._db)
 
-def collection_from_details(_db, client, collection_details):
-    if not (client and isinstance(client, IntegrationClient)):
-        return None
```
leonardr: This method was moved to `Controller.load_collection` -- similar to `CirculationManagerController.load_licensepool` and the like.

controller.py
```diff
@@ -81,46 +83,157 @@
...
+        # and data source name.
+        try:
+            collection, ignore = Collection.from_metadata_identifier(
+                self._db, metadata_identifier, data_source=data_source_name
+            )
+        except ValueError as e:
+            # This is caused by a problem decoding the metadata
+            # identifier.
+            return INVALID_INPUT.detailed(unicode(e))
+        return collection
```
leonardr: Previously, if you passed in a bad metadata identifier, you'd get an exception that didn't explain what went wrong.

controller.py
```diff
@@ -1,7 +1,8 @@
 import base64 as stdlib_base64
 from nose.tools import set_trace
 from datetime import datetime
-from flask import request, make_response
+import flask
```
leonardr: I'm importing flask directly so we can refer to `flask.request` (an established pattern we use to mean "the current request") rather than just `request`.

controller.py
```diff
@@ -805,24 +913,23 @@ class URNLookupController(CoreURNLookupController):
 
     log = logging.getLogger("URN lookup controller")
 
-    def __init__(self, _db, coverage_provider_kwargs=None):
+    def __init__(
+            self, _db, identifier_resolver_class=IdentifierResolutionCoverageProvider,
```
leonardr: I didn't end up swapping out identifier_resolver_class, but I think it'd be useful in the future to test URNLookupController on a more granular basis. The tests we have provide good coverage but they're more high-level than I'd like.

controller.py
```diff
@@ -805,24 +913,23 @@ class URNLookupController(CoreURNLookupController):
...
+        self.identifier_resolver_class = identifier_resolver_class
         self.coverage_provider_kwargs = dict(coverage_provider_kwargs or {})
 
-    @property
-    def default_collection(self):
-        if getattr(self, '_default_collection_id', None) is None:
-            default_collection, ignore = IdentifierResolutionCoverageProvider.unaffiliated_collection(self._db)
-            self._default_collection_id = default_collection.id
-        return get_one(self._db, Collection, id=self._default_collection_id)
-
```
leonardr: This got moved to `Controller`.

controller.py
```diff
@@ -832,7 +939,7 @@ def presentation_ready_work_for(self, identifier):
             return work
         return None
 
-    def process_urns(self, urns, collection_details=None, **kwargs):
+    def process_urns(self, urns, metadata_identifier=None, **kwargs):
```
leonardr: I changed `collection_details` to `metadata_identifier` throughout to make it more specific. I had to also make sure to change this in `routes.py`, where we pass keyword arguments into methods like this one, and our unit tests don't have a way of checking whether the arguments match -- only that the controller has a method with the right name.

tests/test_controller.py
```diff
@@ -80,6 +85,39 @@
...
+        from app import app
+
+        secret = args[0].client.shared_secret.encode('utf8')
+        valid_auth = 'Bearer '+ base64.urlsafe_b64encode(secret)
+        headers = { 'Authorization' : valid_auth }
+        with app.test_request_context('/', headers=headers):
+            return f(*args, **kwargs)
+    return decorated
+
+
```
leonardr: I moved these annotators out of TestIntegrationClientController so they could be used in other tests. They can't completely replace the need for the `authenticated_request` context manager, but when it is possible to use them, the test code is simpler.

controller.py
```diff
@@ -333,8 +444,9 @@ def updates_feed(self, collection_details):
                 precomposed_entries.append(entry)
             else:
                 # There is no cached OPDS entry. We'll create one later.
-                works_for_feed.append((work, identifier))
+                works_for_feed.append((identifier, work))
```
leonardr: (work, identifier) is the wrong order. This error wasn't caught because in all of the test cases, there was already a cached OPDS entry for the work.

tests/test_controller.py
```diff
@@ -236,9 +377,15 @@ def setup(self):
...
 
+        # Create two works to use in tests.
         self.work1 = self._work(with_open_access_download=True)
         self.work2 = self._work(with_open_access_download=True)
 
+        # Clear the cached OPDS entries for one of these works -- this
+        # verifies that OPDS entries are created when needed.
+        self.work1.verbose_opds_entry = None
+        self.work1.simple_opds_entry = None
+
```
leonardr: This guarantees we run through the code path that caused the "wrong order" bug fixed here: https://github.com/NYPL-Simplified/metadata_wrangler/pull/199/files#r314781382

tests/test_controller.py
```diff
@@ -271,31 +418,29 @@ def assert_message(cls, message, identifier, status_code, description):
         eq_(str(status_code), cls.xml_value(message, 'simplified:status_code'))
         eq_(description, cls.xml_value(message, 'schema:description'))
 
+    @unauthenticated_request_context
```
leonardr: Using @unauthenticated_request_context lets us simplify the rest of the code in this test -- we don't have to create test request contexts because the entire test runs inside a single test request context.

tests/test_controller.py
```diff
@@ -621,16 +777,21 @@ def test_metadata_needed_for(self):
             metadata_already_id,
         ])
 
-        with self.app.test_request_context(headers=self.valid_auth):
+        with self.authenticated_request():
             response = self.controller.metadata_needed_for(self.collection.name)
 
-        m = messages = self.get_messages(response.get_data())
+        [m] = self.get_messages(response.get_data())
```
leonardr: The old code wasn't checking that there is only a single message in the response.

tests/test_coverage.py
```diff
@@ -142,7 +145,14 @@ def test_gather_providers_content_cafe(self):
...
         # The OCLC provider was already configured; now there is also a ContentCafeCoverageProvider.
-        provider = IdentifierResolutionCoverageProvider(self._default_collection)
+
+        # Instantiating a real ContentCafeAPI will make an HTTP
+        # requests to get a WSDL file, and we don't want that in a
+        # unit test.
+        provider_kwargs = {
+            ContentCafeCoverageProvider : dict(api=MockContentCafeAPI())
+        }
+        provider = IdentifierResolutionCoverageProvider(self._default_collection, provider_kwargs=provider_kwargs)
```
leonardr: This is unrelated to everything else in this branch. I ran the test suite while eating lunch (no wifi) and discovered that this test was making a real HTTP request to get the WSDL file from Content Cafe.

tests/test_routes.py
```diff
@@ -0,0 +1,346 @@
...
+from core.opds import VerboseAnnotator
+
+from app import app
+from controller import MetadataWrangler
+from problem_details import INVALID_CREDENTIALS
+import routes
+
+from test_controller import ControllerTest
+
+class MockMetadataWrangler(object):
```
leonardr: This code is very similar to [the mock controller scaffolding in circulation](https://github.com/NYPL-Simplified/circulation/blob/master/tests/test_routes.py#L33), but it's not the same because here the application and authentication setup are different (simpler).

controller.py
```diff
@@ -81,46 +83,157 @@
...
+        """
         self._db = _db
+        self._default_collection_id = None
+
+    @property
+    def default_collection(self):
+        """Look up the 'unaffiliated' collection, which is used to register
+        identifiers referenced by unauthenticated clients in one-off tests.
+        """
+        if self._default_collection_id is None:
```
EdwinGuzman: Why was this changed from `getattr(self, ...` to `self._default ...`?

leonardr: _default_collection_id used to be set the first time it was used. Before that happened, `self._default_collection_id` would have raised an exception. Now it's set to `None` in the constructor, so we can always check the value, even if it turns out to be `None`.

tests/test_controller.py
EdwinGuzman: So both of these decorators aren't currently being used yet?

leonardr: They are used, but only on tests where the entire test runs in a single request where the only interesting thing about the request is "it's authenticated".

leonardr:
n.b. We agreed to talk about Flask in person tomorrow if Edwin still has questions.

PR #200, opened 2019-08-16
Add tests for FAST and LCSH loading code, and the FAST name assignment monitor

leonardr:
This branch adds test coverage for fast.py. A lot of the unit tests are more holistic than they could be -- there's a lot of filesystem access, so this was easier to write -- but we now have 97% coverage of this file instead of 0% coverage.

I found and fixed a serious error that would have prevented the LCSH loading from working using the actual LCSH ZIP files we include in the metadata repo. In point of fact, it looks like I loaded LCSH data into our metadata wrangler using some code other than what I committed to the repo -- that's why we didn't notice the problem earlier.

It doesn't show up well in a diff, so here's the directory structure I set up for the tests, underneath `test/files/FAST`:

```
not-consolidated/
 FAST/
  file1.nt.gz
  file2.nt.gz
  random_junk
 LCSH
  test_childrens_subjects_nt.zip
  test_genre_subjects.nt.zip

consolidated/
 FAST/
  broken.nt.gz
  consolidated.csv.gz
```

The "not-consolidated" directory simulates the initial setup. The FAST directory contains gzipped .nt files and the LCSH directory contains ZIP files, each of which contains one .nt file. The "random_junk" file proves that any unrecognized file is ignored.

The "consolidated" directory simulates the situation after initial setup. The data has been loaded from .nt.gz or ZIP files and a consolidated CSV file has been written out. `broken.nt.gz` would crash the loading code if we tried to load it, but we don't bother because we have `consolidated.csv.gz`.

I'm reading real files from disk because it makes test setup much easier. But I don't want a unit test to _write_ files to disk (especially since the presence of a consolidated.csv.gz file will change the behavior of the code), so I mocked the code that writes the consolidated CSV file.

After fixing the underlying loader I was able to fix the bin/subjects_assign script, which uses FAST and LCSH data to assign English names to any FAST or LCSH subjects that have identifiers but not names. This is important because we can only make decisions about fiction status, genre, etc. based on the names -- the identifiers mean nothing to us on their own.

The bin/subjects_assign script has been totally broken for a long time, though it hasn't been a huge deal because we successfully assigned names to all the popular FAST and LCSH subjects a long time ago.


fast.py
```diff
@@ -1,58 +1,130 @@
...
+    def load_triples_file(self, path):
+        """Load classifications from an N-Triples file."""
+        if path.endswith(".nt.gz"):
+            # This is a single GZipped N-Triples file.
+            self.load_triples_filehandle(gzip.open(path, 'rb'))
+        elif path.endswith(".nt.zip"):
+            # This is a ZIP file containing one or more (probably just
+            # one) N-Triples files. Load each one the zip.
+            for fh in self.triples_filehandles_from_zip(path):
+                self.load_triples_filehandle(fh)
```
leonardr: Previously, we handled .nt.gz and .nt files, but ignored .nt.zip files. But .nt.zip is what we actually get from the LCSH dump -- a bunch of ZIP files, each with an .nt file buried deep within its directory structure. I introduced triples_filehandles_from_zip to deal with this.

fast.py
```diff
@@ -1,58 +1,142 @@
...
-            g = self.triple_re.search(triple)
-            if not g:
-                continue
-            identifier, name = g.groups()
-            self[identifier] = name
-
     @classmethod
     def from_data_directory(cls, data_directory):
+        """Load names from a directory that either contains a bunch of
+        files in N-Triples format or a single consolidated CSV file.
```
EdwinGuzman: Are N-Triple files only handled here from this data source in all of the project?

leonardr: Yeah, that's the format used in the data dumps. We deal with RDF data in the OCLC Linked Data code, but it's in JSON-LD format. And even here we're not doing proper N-Triple processing, just scanning for certain regular expressions.

fast.py
```diff
@@ -1,58 +1,142 @@
...
-            triple = triple.strip()
-            g = self.triple_re.search(triple)
-            if not g:
-                continue
-            identifier, name = g.groups()
-            self[identifier] = name
-
     @classmethod
     def from_data_directory(cls, data_directory):
+        """Load names from a directory that either contains a bunch of
```
EdwinGuzman: Is "names" the subject in each line of an N-Triple file? If I understand it correctly, if any row has an identifier but not a name, a name will be assigned?

tests/test_fast.py
```diff
@@ -0,0 +1,91 @@
...
+    def test_from_data_directory_not_consolidated(self):
+        # Load FAST data from a number of ZIP archives containing data
+        # in N-Triples format.
+        not_consolidated = MockLCSHNames.from_data_directory(
+            os.path.join(BASE_DIR, "files/fast/not-consolidated")
+        )
+        eq_(
+            {
+                'sj00001253': 'Ceratopsians',
+                'gf2008025611': 'Abstract films'
```
EdwinGuzman: Is one of the differences between FAST and LSCH that LSCH has letters in the identifiers?

tests/test_monitor.py
```diff
@@ -0,0 +1,73 @@
...
+
+        lcsh1 = self._subject(Subject.LCSH, "lcsh1")
+        lcsh2 = self._subject(Subject.LCSH, "lcsh2")
+        lcsh2.name = "Existing LCSH name."
+
+        # This Subject has a name but no identifier -- this probably
+        # shouldn't happen.
+        missing_identifier = self._subject(Subject.LCSH, "")
+        missing_identifier.identifier = None
+        missing_identifier.name = None
```
EdwinGuzman: The comment says this has a name but here it's set to None. Unless if name is something else?

fast.py
leonardr: Yes, that's right. In our `subjects` table we have an "identifier" and a "name". The "identifier" is usually a number from a controlled vocabulary, like `sj00001253`. This script assigns the English "name" (e.g. "Ceratopsians") to any "identifiers" where we originally found out about the identifier but not the name.

tests/test_fast.py
leonardr: FAST doesn't have letters in the identifiers, but I don't know what the letters mean in LCSH identifiers. They're just different systems, although FAST is based on LCSH.

Looking it up, I see that "sj" means a juvenile subject heading. So it's the faceting information. In some cases we could derive some information from the FAST identifiers that we couldn't get from the names. "sj" means this is a book for kids (though we have no idea how old the kids are).

tests/test_monitor.py
leonardr: The comment was left over from a prior version of this test that didn't test as much. I fixed the comment.

tests/test_fast.py
EdwinGuzman: The regex for each system is also different so it seems they get the headings from different sources.

leonardr: Yeah, they come from different orgs -- I think LSCH comes from Library of Congress and FAST comes from OCLC.

PR #202, opened 2019-11-19
A disabled IntegrationClient can't use an authentication-required service

leonardr:
This branch implements https://jira.nypl.org/browse/SIMPLY-2418 by allowing the metadata wrangler to keep track of certain integration clients that it just won't allow to use the site. The goal here is to stop servers with known problems (such as https://jira.nypl.org/browse/SIMPLY-2266) from bringing down the metadata wrangler.

We don't have an interface for enabling or disabling these clients, but it shouldn't be an issue -- we shouldn't need to do this a whole lot.

### DIFF
--- a/coverage_utils.py
+++ b/coverage_utils.py
@@ -1,3 +1,4 @@
+import logging
 from nose.tools import set_trace
 
 from sqlalchemy.orm.session import Session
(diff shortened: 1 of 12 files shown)


leonardr:
Looks like the core change means I need to remove the assumption of a site-wide mirror integration.

leonardr:
@EdwinGuzman This is derailing the main purpose of this branch, but since you got rid of the "site-wide mirror integration" assumption I want to talk out with you how to deal with the test failures. 

Basically for the metadata wrangler _ think we _do_ want to have a single site-wide mirror integration. That's the situation we have now and there's no need to change it. So I think I'll basically recreate `MirrorUploader.sitewide` inside the metadata wrangler code. Whenever I create a ReplacementPolicy, I'll use the sitewide integration as the cover mirror, and we won't have a book mirror.

Does this make sense?

EdwinGuzman:
Yeah, that strategy makes sense. I think we vaguely talked about how this would affect the MW and now the time has come. Will the database need to be updated as well?

leonardr:
No, neither the schema nor the contents of the database should have to change.

coverage_utils.py
```diff
@@ -10,26 +11,54 @@
...
+            integration = get_one(
+                _db, ExternalIntegration, goal=ExternalIntegration.STORAGE_GOAL
+            )
+            if integration:
+                mirror = MirrorUploader.implementation(integration)
+            else:
+                logging.error(
+                    "No storage integration is configured. Cover images will not be mirrored. You really ought to set up a storage integration."
+                )
+                mirror = None
```
leonardr: In here we keep all the details about how exactly we create `MirrorUploader` objects for a metadata wrangler. Then we just need to make sure we always make a `MetadataWranglerReplacementPolicy` instead of a normal `ReplacementPolicy`. If anything else like this shows up in the future we can put it in this class as well.

integration_client.py
```diff
@@ -143,10 +143,7 @@ def __init__(self, collection, *args, **kwargs):
 
         replacement_policy = kwargs.pop('replacement_policy', None)
         if not replacement_policy:
-            mirror = MirrorUploader.sitewide(_db)
-            replacement_policy = ReplacementPolicy(
-                mirror=mirror, links=True
-            )
```
leonardr: This would have caused test failures when core was updated, but this bit of code wasn't tested. I found it by searching all the code for `ReplacementPolicy`.

tests/test_integration_client.py
```diff
@@ -166,6 +166,22 @@ def setup(self):
             replacement_policy=replacement_policy, collection=self.collection
         )
 
+    def test_default_replacement_policy(self):
```
leonardr: Here's the test that was previously missing.

leonardr:
OK, I think this is good to review now.

coverage_utils.py
```diff
@@ -10,26 +11,54 @@
...
+        """Create a MetadataWranglerReplacementPolicy based on
+        database settings -- specifically, with the configured
+        MirrorUploader for cover images.
+
+        :param mirror: Pass in a mock MirrorUploader to use it
+        instead of creating a real one.
+        """
+        mirror = kwargs.pop('mirror', None)
+        if not mirror:
+            integration = get_one(
```
EdwinGuzman: Won't this cause an issue if there are more than one ExternalIntegrations with the storage goal?

leonardr: On the metadata wrangler, having more than one storage integration _is_ the problem -- we don't currently know what that would mean in real-world terms. If there's no storage integration, then I think the best thing is to log an error and proceed -- the storage integration isn't required. If there are two storage integrations, then I really don't know what should happen, and I think we should stop work until the situation is resolved.

PR #209, opened 2021-02-23
Core update

EdwinGuzman:
This updates the core submodule to the latest on its `main` branch.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,7 +24,7 @@ money
 pymarc
 accept-types
 watchtower # for cloudwatch logging
-pyspellchecker
+pyspellchecker==0.5.6
 
 # nltk is a textblob dependency, and this is the last release that supports Python 2
 nltk==3.4.5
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -24,7 +24,8 @@ money
 pymarc
 accept-types
 watchtower # for cloudwatch logging
-pyspellchecker
+pyspellchecker==0.5.6
+parameterized
```
EdwinGuzman, line 28: `metadata_wrangler` doesn't use this package but `server_core` does. The tests on Travis failed though because it couldn't find this module. I thought it would recursively install all modules from all `requirements` files, no?

leonardr: You can get that behavior by sticking `-r core/requirements.txt` at the top of this requirements.txt file.

leonardr (approved):

requirements.txt
EdwinGuzman: Thanks. I'll leave it like this for now because there are a lot of overlapping modules which will most likely get updated with the python 3 update.

PR #211, opened 2021-02-25
Python 3 conversion

EdwinGuzman:
~I used 2to3 to convert some of the code but kept unicode instead of str in places where I didn't understand which one was correct and also to pass the current set of tests.
Also added a PR template file and updated `server_core` to use the `develop` branch, so now the packages in `requirements.txt` are better aligned to core and also have specific versions.
For the next step, I'm thinking of using tox to make sure tests pass in python2.7 and python 3.6 and then fully convert to python 3. Or, does it make more sense to just update completely to python 3?~

March 29th update:
* This PR branch now depends on the python 3 branch for `server_core` and moves to only support python 3 (3.6 to 3.9).
* `tox` is added to manage the different python testing environments.
* Switched from using Travis for CI/CD to Github Actions.
* Switched from using nose to pytest.


app.py
```diff
@@ -80,8 +80,8 @@ def shutdown_session(exception):
 from routes import *
 
 def run(self, url=None, debug=False):
-    base_url = url or u'http://localhost:5500/'
-    scheme, netloc, path, parameters, query, fragment = urlparse.urlparse(base_url)
+    base_url = url or 'http://localhost:5500/'
```
leonardr, line 83: My only real concern here is stuff like this where we lose the Unicode marker in a string. If we're not moving right to Python 3 I'd like to keep these markers.

It's tough for me to say whether any particular case matters, because we don't have great test coverage of this code base.

EdwinGuzman:
That's a good plan. I will keep this PR open until the full conversion/docker update.

fast.py
```diff
@@ -124,16 +124,16 @@ def write_consolidated_file(self, path):
...
 
     @contextmanager
     def consolidated_output_filehandle(self, path):
         """Open a write filehandle to the given path.
-
         This method is designed to be mocked in unit tests.
         """
-        with gzip.open(path, "wb") as out:
+        with gzip.open(path, "w") as out:
+            set_trace()
```
leonardr, line 136: Test code that should be removed.

EdwinGuzman:
The last commit:
* added tox to test with different python environments
* removed nose and added pytest
* removed the travis.yml file and added a test file to run tests in Github Actions

PR #213, opened 2021-04-08
UTC DateTime update

EdwinGuzman:
This is part of the python 3 conversion but has a specific `core` branch that converts datetime objects into timezone-aware datetime objects with UTC information. This points to this `server_core` PR https://github.com/NYPL-Simplified/server_core/pull/1256 and will be ready once that PR is merged.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -65,6 +65,7 @@ from core.util.http import HTTP
 from core.util.opds_writer import OPDSMessage
 from core.util.problem_detail import ProblemDetail
 from core.util.string_helpers import base64
+from core.util.datetime_helpers import strptime_utc
 from core.util.flask_util import OPDSFeedResponse
 
 from coverage_provider import (
@@ -416,7 +417,7 @@ class CatalogController(Controller):
         last_update_time = flask.request.args.get('last_update_time', None)
         if last_update_time:
             try:
-                last_update_time = datetime.strptime(
+                last_update_time = strptime_utc(
                     last_update_time, self.TIMESTAMP_FORMAT
                 )
             except ValueError as e:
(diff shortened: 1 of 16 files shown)


controller.py
```diff
@@ -416,7 +417,7 @@ def updates_feed(self, metadata_identifier):
         last_update_time = flask.request.args.get('last_update_time', None)
         if last_update_time:
             try:
-                last_update_time = datetime.strptime(
+                last_update_time = strptime_utc(
                     last_update_time, self.TIMESTAMP_FORMAT
```
leonardr, line 421: While looking at this I discovered that my core branch is wrong. The  "Z" in the TIMESTAMP_FORMAT doesn't describe a timezone-naive datetime, although the Atom RFC seems to say that's what it's for. The "Z" means Zulu or Zero time, i.e. UTC. That's what I thought originally.
 
I was going to say we need to get rid of the "Z" in timestamps whenever we're communicating across machines, but this is correct. The flip side is that if we do get a naive datetime there's no way to represent that in Atom -- best we can do is pretend it's UTC.

leonardr (approved):

controller.py
EdwinGuzman: When you say "but this is correct", do you mean that leaving the "Z" in the timestamp is correct?

Looking at this again, I just realized that `TIMESTAMP_FORMAT` is missing `%z` in the string. That's what picks up the UTC offset in an ISO-8601 date string. The "Z" should be equivalent to an UTC offset so I think it's equivalent to have either "Z" or "+00:00", right? If we do leave the "Z", I think it's the default +00:00 without an offset but _that_ may be wrong.

leonardr: Yeah, leaving "Z" in TIMESTAMP_FORMAT is correct _if_ the time is UTC. Otherwise you have to have a numeric offset (or, I guess, omit timezone information altogether).

`Z` is equivalent to `+00:00` and a given timestamp should only have one of them.

PR #214, opened 2021-04-15
Docker update

EdwinGuzman:
Resolves [SIMPLY-3541](https://jira.nypl.org/browse/SIMPLY-3541).

This is branched off the `datetime-update` branch.

This PR implements the minimum to get Docker to run the application in python 3. This also updates the Postgres database from 9.5 to 12 in Docker and in tox that runs pytests.

I updated the README regarding running the application locally with and without Docker. One limitation of using Docker is that it's meant for deployment only and not for local development. Since I just wanted to get this working with python 3 and not add any new features, this would be great to look into in the future. It may be as simple as mounting the current directory with the directory the webapp docker container uses.


leonardr (approved):
I filed SIMPLY-3706 to cover further improvements to the Dockerfile and the CI process; this is plenty for now.

PR #215, opened 2021-04-15
Python 3 Update

EdwinGuzman:
This is the same PR as PR #211 except that this now points to the `develop`  branch. I removed the `master` branch before making the switch and so the original PR closed automatically and can't be edited.

Check that PR for more information but it basically includes:
This PR branch now depends on the python 3 branch for server_core and moves to only support python 3 (3.6 to 3.9).
* tox is added to manage the different python testing environments.
* Switched from using Travis for CI/CD to Github Actions.
* Switched from using nose to pytest.


leonardr (approved):
Good thing I reviewed the other PR!

